PR #126, opened 2025-07-14
Docc pipelining article

adam-fowler:
Also fixed up doc link in function header


README.md
```diff
@@ -19,30 +19,32 @@ try await withThrowingTaskgroup(of: Void.self) { group in
...
 In some cases it is desirable to send multiple commands at one time, without waiting for the response after each command. This is called pipelining. You can do this using the function `pipeline(_:)`. This takes a parameter pack of commands and returns a parameter pack with the responses once all the commands have executed.
 
 ```swift
-let (setResponse, getResponse) = try await connection.pipeline(
+let (setResponse, getResponse) = await connection.pipeline(
     SET(key: "MyKey", value: "TestString"),
     GET(key: "MyKey")
 )
+let value = try getResponse.get()
 ```
```
fabianfett, line 48: we should also call out that you can also pipeline on a connection using `async let`. It just won't be as efficient.

Sources/Valkey/Documentation.docc/Pipelining.md
```diff
@@ -0,0 +1,43 @@
...
+
+Sending multiple commands at once without waiting for the response of each command
+
+## Overview
+
+Valkey pipelining is a technique for improving performance by issuing multiple commands at once without waiting for the response to each individual command. Pipelining not only reduces the latency cost of waiting for the result of each command it also reduces the cost to the server as it reduces I/O costs. Multiple commands can be read with a single syscall, and multiple results are delivered with a single syscall. 
+
+## Implementation
+
+In valkey-swift each command has its own type conforming to the protocol ``ValkeyCommand``. This type has an `associatedtype` ``ValkeyCommand/Response``. The ``ValkeyClient/pipeline(_:)`` command takes a parameter pack of types conforming to ``ValkeyCommand`` and returns a parameter pack containing the result of each command.
```
fabianfett, line 11:
```suggestion
In valkey-swift each command has its own type conforming to the protocol ``ValkeyCommand``. This type has an `associatedType` ``ValkeyCommand/Response``. The ``ValkeyClient/pipeline(_:)`` command takes a parameter pack of types conforming to ``ValkeyCommand`` and returns a parameter pack containing the result of each command.
```

README.md
```diff
@@ -19,30 +19,32 @@ try await withThrowingTaskgroup(of: Void.self) { group in
...
 
 ```swift
-try await valkeyClient.withConnection { connection in
-    try await doValkeyStuff()
-}
+try await valkeyClient.set(key: "foo", value: "bar")
 ```
-
-All the Valkey commands are in the Commands folder of the Valkey target. These are generated from the model files Valkey supplies in [valkey](https://github.com/valkey-io/valkey/src/commands). In many cases where it was possible to ascertain the return type of a command these functions will return that expected type. In situations where this is not possible (or we are returning a String) a `RESPToken` is returned and you'll need to convert it manually.
+Or you can create a connection and run multiple commands from that connection using `ValkeyClient.withConnection()`.
```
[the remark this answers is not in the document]
adam-fowler: Not sure why you would mention it here, as well

README.md
adam-fowler: I think we should leave the README to be fairly minimal and expand on stuff in the docc documentation

fabianfett: agree. fine if we skip it here.

PR #140, opened 2025-07-17
Add `@discardableResult` to commands that are not readonly

adam-fowler:
The logic being these commands write to the database somehow, so you are probably calling them for the change to the database and not the return type. eg `SET`, `LPUSH`.


fabianfett (approved):
LGTM

PR #143, opened 2025-07-18
Spelling

jsoref:
I've [skipped a number of things that upset the API validator](https://github.com/jsoref/valkey-swift/compare/f7468fc08bd02ee98f9114a69985546b84bb0752..6f777821fc0b87de71951527d041ddfd7192d03f) even though I really would want to fix them


adam-fowler (commented):
Thanks for these
Some of the code is generated or is built from sources elsewhere. You will lose changes in the `Sources/Valkey/Commands` folder as the file in the root of that folder are generated. You will also lose the changes to the json files in `_ValkeyCommandsBuilder/Resources`.

The source for these can be found in https://github.com/valkey-io/valkey/tree/unstable/src/commands and https://github.com/valkey-io/valkey-json/tree/unstable/src/commands

adam-fowler (approved):

adam-fowler:
Thank you for this

PR #148, opened 2025-07-22
adds public api docs

heckj:
- **adds docs to public API that was otherwise missing**
- **updates based on feedback in earlier PR**


heckj:
@fabianfett Your suggested updates should all be applied, primarily in the second commit (https://github.com/valkey-io/valkey-swift/pull/148/commits/d3bb8bd012f0199dd320e7367c55d40993192b35) - if that makes it easier to look through.

adam-fowler (approved):

heckj:
rebased, fixed the conflict resolution, and applied formatting

PR #158, opened 2025-07-25
Make ValkeyClient go brrrr

fabianfett:
### Motivation

ConnectionPool offers a fast mode for single requests. For this to work we have to leave the paradigms of structured concurrency. It is important that the ConnectionRequest is synchronously fulfilled without any continuations.

Using the `withConnection` API, we have the following context switch scenario:

- On EL, a response has been received
- On EL, the commands continuation is succeeded
- JUMP to concurrent executor, EL is now potentially idle
- On CE (concurrent executor) end of `withConnection` is reached -> acquire lock -> get next waiting request -> fulfill continuation for connection
- On CE next `withConnection` closure is invoked
- On CE next `send` is invoked
- JUMP to EL for actor isolation, we are back on the EL and can run the command

If we don't use structured concurrency here, we can have the following flow:

- On EL, a response has been received
- On EL, we can release the connection -> acquire lock -> get next waiting request -> invoke request callback with connection
- On EL request callback is invoked and can write command right away

Not a single context switch was necessary in this scenario.

### Performance

In local testing I can see about a 20% improvement (throughput and wall clock time) for all the client benchmarks.

### Changes

- Add `ValkeyConnectionRequest`, that uses unstructured callbacks for improved performance

### DIFF
--- a/Sources/Valkey/Connection/ValkeyConnection.swift
+++ b/Sources/Valkey/Connection/ValkeyConnection.swift
@@ -36,7 +36,9 @@ public final actor ValkeyConnection: ValkeyConnectionProtocol, Sendable {
     /// Logger used by Server
     let logger: Logger
     @usableFromInline
-    let channel: any Channel
+    nonisolated let executor: any NIOSerialEventLoopExecutor
+    @usableFromInline
+    nonisolated let channel: any Channel
     @usableFromInline
     let channelHandler: ValkeyChannelHandler
     let configuration: ValkeyConnectionConfiguration
@@ -50,7 +52,11 @@ public final actor ValkeyConnection: ValkeyConnectionProtocol, Sendable {
         configuration: ValkeyConnectionConfiguration,
         logger: Logger
     ) {
-        self.unownedExecutor = channel.eventLoop.executor.asUnownedSerialExecutor()
+        guard let executor = channel.eventLoop as? any NIOSerialEventLoopExecutor else {
+            fatalError()
+        }
+        self.executor = executor
+        self.unownedExecutor = executor.asUnownedSerialExecutor()
         self.channel = channel
         self.channelHandler = channelHandler
         self.configuration = configuration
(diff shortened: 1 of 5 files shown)


Sources/Valkey/Connection/ValkeyConnection.swift
```diff
@@ -261,10 +282,13 @@ public final actor ValkeyConnection: ValkeyConnectionProtocol, Sendable {
...
         }
 
+
+
         let future: EventLoopFuture<Channel>
         switch address.value {
         case .hostname(let host, let port):
-            future = connect.connect(host: host, port: port)
+            let socketAddress = try! SocketAddress(ipAddress: host, port: port)
+            future = connect.connect(to: socketAddress)
```
adam-fowler, line 291: Can see why you did this, but can we keep it to another PR. It would be good to still support hostnames.

Sources/Valkey/ValkeyClient.swift
```diff
@@ -209,3 +224,112 @@ extension ValkeyClient {
...
+    typealias Connection = ValkeyConnection
+
+    @usableFromInline
+    let id: Int
+    @usableFromInline
+    let pool: ValkeyClient.Pool
+    @usableFromInline
+    let continuation: CheckedContinuation<T, any Error>
+    @usableFromInline
+    let lock: Mutex<RequestState>
```
adam-fowler, line 270: You setup RequestState to be AtomicRepresentable, but are using a Mutex

Sources/Valkey/ValkeyClient.swift
```diff
@@ -148,19 +153,20 @@ extension ValkeyClient {
         isolation: isolated (any Actor)? = #isolation,
         operation: (ValkeyConnection) async throws -> sending Value
     ) async throws -> Value {
-        let connection = try await self.leaseConnection()
-
-        defer { self.connectionPool.releaseConnection(connection) }
-
-        return try await operation(connection)
+        fatalError()
```
adam-fowler, line 156: And how are you planning to support withConnection now that the connection request is `ValkeyConnectionRequest<RESPToken>`

Sources/Valkey/ValkeyClient.swift
```diff
@@ -209,3 +224,112 @@ extension ValkeyClient {
...
+    }
+
+    @inlinable
+    func complete(with result: Result<Connection, ConnectionPoolError>) {
+        switch result {
+        case .success(let connection):
+            self.lock.withLock { state in
+                state = .onConnection(connection)
+            }
+            self.onConnection(connection, self)
```
adam-fowler, line 295: complete and succeed seem to do different things for a successful result, which is the correct one?

PR #170, opened 2025-07-31
Call CLIENT SETINFO on initialisation of connection

adam-fowler:
This also adds a new fire and forget API along with a ValkeyPromise type `.forget` where we don't care about the result (to avoid extra allocations).

### DIFF
--- a/Sources/Valkey/Connection/ValkeyConnection.swift
+++ b/Sources/Valkey/Connection/ValkeyConnection.swift
@@ -146,8 +146,8 @@ public final actor ValkeyConnection: ValkeyClientProtocol, Sendable {
 
     func initialHandshake() async throws {
         try await self.channelHandler.waitOnActive().get()
-        self.executeAndForget(command: CLIENT.SETINFO(attr: .libname("valkey-swift")))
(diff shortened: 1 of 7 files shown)


Sources/Valkey/Connection/ValkeyConnection.swift
```diff
@@ -144,8 +144,10 @@ public final actor ValkeyConnection: ValkeyClientProtocol, Sendable {
         self.channel.close(mode: .all, promise: nil)
     }
 
-    func waitOnActive() async throws {
+    func initialHandshake() async throws {
         try await self.channelHandler.waitOnActive().get()
+        self.executeAndForget(command: CLIENT.SETINFO(attr: .libname("valkey-swift")))
+        self.executeAndForget(command: CLIENT.SETINFO(attr: .libver("0.1.0")))
```
fabianfett, line 150: we should get this string from a static version string, that we maintain in `Version.swift` or whatever. Before every release we should bump it.

adam-fowler:
We kind of are in that we don't wait for the response from either. We could send them as one buffer though which we aren't doing at the moment.

fabianfett:
yeah my point was about reducing sys calls here, which should make connection creation even faster.

PR #171, opened 2025-08-01
Add pipelining benchmark

adam-fowler:
I know of some optimisations we can make to pipelining, this adds a benchmark so we can test them.


fabianfett (approved):
LGTM

PR #172, opened 2025-08-03
Move SETINFO calls to ChannelHandler so they can be written along with HELLO

adam-fowler:


Sources/Valkey/Connection/ValkeyChannelHandler+stateMachine.swift
```diff
@@ -65,11 +65,12 @@ extension ValkeyChannelHandler {
...
-            var pendingHelloCommand: PendingCommand
+            let pendingHelloCommand: PendingCommand
+            var pendingCommands: Deque<PendingCommand>
 
-            func cancel(requestID: Int) -> PendingCommand? {
+            func cancel(requestID: Int) -> [PendingCommand]? {
                 if pendingHelloCommand.requestID == requestID {
-                    return pendingHelloCommand
+                    return [pendingHelloCommand] + pendingCommands
                 }
```
fabianfett, line 74: I think the cancel code here should be unreachable. HELLO and SETINFO are triggered from within the channel handler.

Sources/Valkey/Connection/ValkeyChannelHandler.swift
```diff
@@ -307,25 +280,35 @@ final class ValkeyChannelHandler: ChannelInboundHandler {
...
+        helloCommand.encode(into: &self.encoder)
+        clientInfoLibName.encode(into: &self.encoder)
+        clientInfoLibVersion.encode(into: &self.encoder)
 
         let promise = eventLoop.makePromise(of: RESPToken.self)
+
         let deadline = .now() + self.configuration.commandTimeout
-        context.writeAndFlush(self.wrapOutboundOut(buffer), promise: nil)
+        context.writeAndFlush(self.wrapOutboundOut(self.encoder.buffer), promise: nil)
         scheduleDeadlineCallback(deadline: deadline)
```
fabianfett, line 303:
```suggestion
        self.scheduleDeadlineCallback(deadline: deadline)
```

fabianfett (commented):
I'm afraid this pr shows, that special casing the HELLO command in the state machine, is not the absolute worst idea, even if it initially adds a bit of code.

Focusing on this approach, I would love to see a "ValkeyConnection create and drop" benchmark for it. 

Generally looks okay, though.

adam-fowler:
> I'm afraid this pr shows, that special casing the HELLO command in the state machine, is not the absolute worst idea, even if it initially adds a bit of code.

This PR shows that having all the commands from the  initial handshake in one place is the best thing. I could have pipelined the hello, client setinfo inside the connection as well.

adam-fowler:
> Can we land the benchmark first and then change the behavior?

See #183

adam-fowler:
> Have you run the tests locally to look at timings? I think that might be a better indicator here?
 
Locally numbers are slightly less erratic, but there are still cases where main is faster
```
╒══════════════════════════════════════════╤═══════════╤═══════════╤═══════════╤═══════════╤═══════════╤═══════════╤═══════════╤═══════════╕
│             Malloc (total) *             │        p0 │       p25 │       p50 │       p75 │       p90 │       p99 │      p100 │   Samples │
╞══════════════════════════════════════════╪═══════════╪═══════════╪═══════════╪═══════════╪═══════════╪═══════════╪═══════════╪═══════════╡
│                   dev                    │       210 │       237 │       239 │       245 │       250 │       250 │       250 │         6 │
├──────────────────────────────────────────┼───────────┼───────────┼───────────┼───────────┼───────────┼───────────┼───────────┼───────────┤
│               Current_run                │       196 │       205 │       210 │       224 │       256 │       256 │       256 │         6 │
├──────────────────────────────────────────┼───────────┼───────────┼───────────┼───────────┼───────────┼───────────┼───────────┼───────────┤
│                    Δ                     │       -14 │       -32 │       -29 │       -21 │         6 │         6 │         6 │         0 │
├──────────────────────────────────────────┼───────────┼───────────┼───────────┼───────────┼───────────┼───────────┼───────────┼───────────┤
│              Improvement %               │         7 │        14 │        12 │         9 │        -2 │        -2 │        -2 │         0 │
╘══════════════════════════════════════════╧═══════════╧═══════════╧═══════════╧═══════════╧═══════════╧═══════════╧═══════════╧═══════════╛

```
All other indicators this PR is considerably faster. ie by 10-20%

fabianfett (approved):

PR #174, opened 2025-08-04
Add ability to client to run other workloads than the connection manager

adam-fowler:
- Add protocol ActionRunner which runs actions inside discarding task group
- Conform ValkeyClient to ActionRunner
- Add single action `runClient` which runs a `ValkeyNodeClient`


adam-fowler:
Don't know if the protocol is overkill, but it could also be used for the cluster client

adam-fowler:
> Would you mind outlining your motivation for this change? I'm sure this will help me understand this change.

To support replica discovery, running replicas connection pools and failover/redirect on a standalone client.

adam-fowler:
> I get that we need to run the ValkeyNodeClients now that they are in the ValkeyClient. However I don't get the motivation for the ActionRunner protocol.

Ok sorry the motivation was to use this also in ValkeyClusterClient and reduce the duplication of code. I do have another branch where it is also used in ValkeyClusterClient.

PR #177, opened 2025-08-05
Add Distributed Tracing support

### LINKED ISSUE #176: Distributed Tracing Support
Would you be open to built-in support for `swift-distributed-tracing`, perhaps behind a trait like the `swift-service-lifecycle` support?

I started looking into what other valkey clients are offering in terms of tracing:

- `valkey-go` has built-in OTel support (https://github.com/valkey-io/valkey-go/blob/main/valkeyotel/trace.go)
- `valkey-py` documents using the existing OTel Redis Python instrumentation library (https://valkey-py.readthedocs.io/en/latest/opentelemetry.html#opentelemetry-instrumentation)

A quick glance at both shows that `valkey-go`'s built-in tracing uses `db.system: valkey` whereas I'd expect the Python instrumentation to use `db.system: redis` given they suggest using the existing Redis instrumentation.

I'd be happy to take a stab at this if you're generally interested in adding Distributed Tracing support.

slashmo:
## Motivation

To make the most of Distributed Tracing, users may want to view operations performed by their database clients as part of a larger request to gain more insight into their distributed systems.

## Modifications

- Add new `DistributedTracingSupport` trait, enabled by default
- Create client span in single and pipelined command execution
- Updated the command builder to generate a command name property, used in the `db.operation.name` span attribute
- Add `open-telemetry` example to showcase Distributed Tracing support

When implementing the span and span attributes I closely followed the OpenTelemetry semantic conventions for Redis and general database client spans:

- https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/redis.md
- https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-spans.md

I believe that this provides the highest chance of the span being recognized as a database client operation by various observability backends.

## Result

Users can now use Distributed Tracing spans automatically created by the library to more deeply observe their systems.

Closes #176

### DIFF
--- a/Sources/Valkey/Connection/ValkeyConnection.swift
+++ b/Sources/Valkey/Connection/ValkeyConnection.swift
@@ -210,7 +210,10 @@ public final actor ValkeyConnection: ValkeyClientProtocol, Sendable {
             throw error
         } catch {
             #if DistributedTracingSupport
-            span?.recordError(error)
+            if let span {
+                span.recordError(error)
+                span.setStatus(SpanStatus(code: .error))
+            }
             #endif
             throw error
         }
--- a/Tests/ValkeyTests/ValkeyConnectionTests.swift
+++ b/Tests/ValkeyTests/ValkeyConnectionTests.swift
@@ -575,6 +575,48 @@ struct ConnectionTests {
             #expect(span.status?.code == .error)
         }
 
+        @Test
+        @available(valkeySwift 1.0, *)
+        func testSingleCommandUnknownFailureSpan() async throws {
+            let tracer = InMemoryTracer()
+            var config = ValkeyConnectionConfiguration()
+            config.tracing.tracer = tracer
+
+            let channel = NIOAsyncTestingChannel()
+            let logger = Logger(label: "test")
+            let connection = try await ValkeyConnection.setupChannelAndConnect(channel, configuration: config, logger: logger)
+            try await channel.processHello()
+
+            async let fooResult = connection.get("foo")
+            _ = try await channel.waitForOutboundWrite(as: ByteBuffer.self)
+
+            await #expect(throws: RESPParsingError.self) {
+                try await channel.writeInbound(ByteBuffer(string: "invalid resp token"))
+            }
+            do {
+                _ = try await fooResult
+                Issue.record()
+            } catch is RESPParsingError {}
+
+            #expect(tracer.finishedSpans.count == 1)
+            let span = try #require(tracer.finishedSpans.first)
+            #expect(span.operationName == "GET")
+            #expect(span.kind == .client)
+            #expect(span.errors.count == 1)
+            let error = try #require(span.errors.first)
+            #expect((error.error as? RESPParsingError)?.code == .invalidLeadingByte)
+            #expect(
+                span.attributes == [
+                    "db.system.name": "valkey",
+                    "db.operation.name": "GET",
+                    "server.address": "127.0.0.1",
+                    "network.peer.address": "127.0.0.1",
+                    "network.peer.port": 6379,
+                ]
+            )
+            #expect(span.status?.code == .error)
+        }
+
         @Test(.disabled("Pipeline support not implemented yet"))
         @available(valkeySwift 1.0, *)
         func testPipelinedSameCommandsSpan() async throws {


Sources/Valkey/Connection/ValkeyConnection.swift
```diff
@@ -161,16 +176,48 @@ public final actor ValkeyConnection: ValkeyClientProtocol, Sendable {
 
     @inlinable
     func _execute<Command: ValkeyCommand>(command: Command) async throws -> RESPToken {
+        #if DistributedTracingSupport
+        let span = startSpan(Command.name, ofKind: .client)
+        defer { span.end() }
```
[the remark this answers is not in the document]
slashmo: Thanks for the suggestion. This is similar to what I originally went with. I eventually switched to the more imperative API because I thought it was easier to read. I'm happy to explore `withSpan` again though.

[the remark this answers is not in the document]
czechboy0: I'll let Adam/Fabian steer you here, I'm also happy with either.

slashmo:
The API breakage check rightfully complains about the added protocol requirement `ValkeyCommand.name`. Is this acceptable given we're pre-1.0 or would you like me to add a default in an extension, for example like this:

```swift
extension ValkeyCommand {
    public static var name: String { String(describing: Self.self) }
}
```

adam-fowler:
> The API breakage check rightfully complains about the added protocol requirement `ValkeyCommand.name`. Is this acceptable given we're pre-1.0 or would you like me to add a default in an extension, for example like this:
> 
> 
> 
> ```swift
> 
> extension ValkeyCommand {
> 
>     public static var name: String { String(describing: Self.self) }
> 
> }
> 
> ```

API breakages are fine

Sources/Valkey/Connection/ValkeyConnection.swift
```diff
@@ -40,6 +44,8 @@ public final actor ValkeyConnection: ValkeyClientProtocol, Sendable {
     @usableFromInline
     let channelHandler: ValkeyChannelHandler
     let configuration: ValkeyConnectionConfiguration
+    @usableFromInline
+    let address: (hostOrSocketPath: String, port: Int?)?
```
adam-fowler, line 48: If port isn't set you can default to 6379

Sources/Valkey/Connection/ValkeyConnection.swift
```diff
@@ -189,8 +236,42 @@ public final actor ValkeyConnection: ValkeyClientProtocol, Sendable {
     public func execute<each Command: ValkeyCommand>(
         _ commands: repeat each Command
     ) async -> sending (repeat Result<(each Command).Response, Error>) {
+        #if DistributedTracingSupport
+        let span = startSpan("MULTI", ofKind: .client)
```
adam-fowler, line 240: Why `MULTI`? This implies you are running the Valkey command `MULTI` which starts a transaction.
Is it worthwhile building a string out of all of the commands eg `SET,LPUSH,INCR`

Tests/ValkeyTests/Utils/TestTracer.swift
adam-fowler, line 1: I wrote a TestTracer that allowed you to run tests in parallel in Hummingbird. https://github.com/hummingbird-project/hummingbird/blob/main/Tests/HummingbirdTests/TracingTests.swift. It used Task Local variables to separate tracers for each test. 

You could probably use a Testing trait to set it up for each test.

Examples/open-telemetry/Package.swift
adam-fowler, line 1: Can we skip the Example just now

slashmo: Sure, I'll remove it.

slashmo:
Here's the benchmark diff between `main` and the pr. I ran them with Swift 6.2, let me know if you'd prefer 6.1 and I can install it:

```console
swift-driver version: 1.127.10 Apple Swift version 6.2 (swiftlang-6.2.0.14.8 clang-1700.3.14.6)
Target: arm64-apple-macosx15.0
```

```console
===============================================================
Threshold deviations for ValkeyBenchmarks:Client: GET benchmark
===============================================================
╒══════════════════════════════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╕
│ Malloc (total) (K, %)                    │            main │              pr │    Difference % │     Threshold % │
╞══════════════════════════════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╡
│ p25                                      │              74 │              81 │              10 │               5 │
├──────────────────────────────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┤
│ p50                                      │              77 │              83 │               8 │               5 │
├──────────────────────────────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┤
│ p75                                      │              78 │              85 │               8 │               5 │
╘══════════════════════════════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╛

=========================================================================================================
Threshold deviations for ValkeyBenchmarks:Client: GET benchmark | parallel 20 | 20 concurrent connections
=========================================================================================================
╒══════════════════════════════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╕
│ Malloc (total) (K, %)                    │            main │              pr │    Difference % │     Threshold % │
╞══════════════════════════════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╡
│ p75                                      │             117 │             125 │               6 │               5 │
╘══════════════════════════════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╛

=========================================================================================================
Threshold deviations for ValkeyBenchmarks:Client: GET benchmark | parallel 50 | 20 concurrent connections
=========================================================================================================
╒══════════════════════════════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╕
│ Malloc (total) (K, %)                    │            main │              pr │    Difference % │     Threshold % │
╞══════════════════════════════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╡
│ p25                                      │               5 │               7 │              30 │               5 │
├──────────────────────────────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┤
│ p50                                      │               8 │              11 │              45 │               5 │
├──────────────────────────────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┤
│ p75                                      │              11 │              18 │              59 │               5 │
╘══════════════════════════════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╛

╒══════════════════════════════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╕
│ Instructions (M, %)                      │            main │              pr │    Difference % │     Threshold % │
╞══════════════════════════════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╡
│ p25                                      │             247 │             269 │               8 │               5 │
├──────────────────────────────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┤
│ p50                                      │             249 │             271 │               9 │               5 │
├──────────────────────────────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┤
│ p75                                      │             250 │             272 │               9 │               5 │
╘══════════════════════════════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╛

===================================================================
Threshold deviations for ValkeyBenchmarks:Connection: GET benchmark
===================================================================
╒══════════════════════════════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╕
│ Throughput (# / s) (#, %)                │            main │              pr │    Difference % │     Threshold % │
╞══════════════════════════════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╡
│ p25                                      │              29 │              27 │               6 │               5 │
├──────────────────────────────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┤
│ p50                                      │              29 │              27 │               6 │               5 │
├──────────────────────────────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┤
│ p75                                      │              29 │              27 │               6 │               5 │
╘══════════════════════════════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╛

╒══════════════════════════════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╕
│ Time (wall clock) (ms, %)                │            main │              pr │    Difference % │     Threshold % │
╞══════════════════════════════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╡
│ p25                                      │              34 │              37 │               6 │               5 │
├──────────────────────────────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┤
│ p50                                      │              35 │              37 │               6 │               5 │
├──────────────────────────────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┤
│ p75                                      │              35 │              38 │               8 │               5 │
╘══════════════════════════════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╛

╒══════════════════════════════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╕
│ Time (total CPU) (ms, %)                 │            main │              pr │    Difference % │     Threshold % │
╞══════════════════════════════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╡
│ p25                                      │              29 │              31 │               6 │               5 │
├──────────────────────────────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┤
│ p50                                      │              29 │              31 │               6 │               5 │
├──────────────────────────────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┤
│ p75                                      │              30 │              32 │               9 │               5 │
╘══════════════════════════════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╛

╒══════════════════════════════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╕
│ Instructions (M, %)                      │            main │              pr │    Difference % │     Threshold % │
╞══════════════════════════════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╡
│ p25                                      │             225 │             247 │               9 │               5 │
├──────────────────────────────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┤
│ p50                                      │             225 │             247 │               9 │               5 │
├──────────────────────────────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┤
│ p75                                      │             225 │             247 │               9 │               5 │
╘══════════════════════════════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╛

========================================================================
Threshold deviations for ValkeyBenchmarks:Connection: Pipeline benchmark
========================================================================
╒══════════════════════════════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╕
│ Throughput (# / s) (#, %)                │            main │              pr │    Difference % │     Threshold % │
╞══════════════════════════════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╡
│ p25                                      │              17 │              16 │               5 │               5 │
├──────────────────────────────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┤
│ p50                                      │              17 │              16 │               5 │               5 │
╘══════════════════════════════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╛

╒══════════════════════════════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╕
│ Malloc (total) (K, %)                    │            main │              pr │    Difference % │     Threshold % │
╞══════════════════════════════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╡
│ p25                                      │              14 │              26 │              83 │               5 │
╘══════════════════════════════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╛

====================================================================
Threshold deviations for ValkeyBenchmarks:HashSlot – {user}.whatever
====================================================================
╒══════════════════════════════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╕
│ Throughput (# / s) (#, %)                │            main │              pr │    Difference % │     Threshold % │
╞══════════════════════════════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╡
│ p75                                      │              57 │              53 │               7 │               5 │
╘══════════════════════════════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╛

╒══════════════════════════════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╕
│ Time (wall clock) (ms, %)                │            main │              pr │    Difference % │     Threshold % │
╞══════════════════════════════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╡
│ p75                                      │              18 │              19 │               7 │               5 │
╘══════════════════════════════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╛

=====================================================================================
Threshold deviations for ValkeyBenchmarks:ValkeyCommandEncoder – Command with 7 words
=====================================================================================
╒══════════════════════════════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╕
│ Time (wall clock) (μs, %)                │            main │              pr │    Difference % │     Threshold % │
╞══════════════════════════════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╡
│ p75                                      │             420 │             441 │               5 │               5 │
╘══════════════════════════════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╛

╒══════════════════════════════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╕
│ Time (total CPU) (μs, %)                 │            main │              pr │    Difference % │     Threshold % │
╞══════════════════════════════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╡
│ p75                                      │             421 │             443 │               5 │               5 │
╘══════════════════════════════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╛

New baseline 'pr' is WORSE than the 'main' baseline thresholds.

error: benchmarkThresholdRegression
```

Tests/ValkeyTests/Utils/TestTracer.swift
```diff
@@ -0,0 +1,209 @@
...
+
+#if DistributedTracingSupport
+import Dispatch
+import Foundation
+import Instrumentation
+import ServiceContextModule
+import Tracing
+
+/// Only intended to be used in single-threaded testing.
+final class TestTracer {
```
fabianfett, line 37: I'm pretty sure this should live in DistributedTracing in a separate module called `TestUtils`. Also this should be proper Sendable from the beginning. Adding annotations to use `Mutex` is fine.

Sources/Valkey/Connection/ValkeyConnection.swift
```diff
@@ -161,16 +176,48 @@ public final actor ValkeyConnection: ValkeyClientProtocol, Sendable {
 
     @inlinable
     func _execute<Command: ValkeyCommand>(command: Command) async throws -> RESPToken {
+        #if DistributedTracingSupport
+        let span = startSpan(Command.name, ofKind: .client)
```
fabianfett, line 180: We should 100% inject a tracer into the ValkeyConnection: `let tracer: (any Tracer)?` We should never go through the global. We might want to get the global when we create the Connection.

fabianfett (commented):
@slashmo Thank you for taking a stab at this one! Really like the direction. We will need to dive deep on some of the aspects.

Tests/ValkeyTests/Utils/TestTracer.swift
adam-fowler: Yes that'd be fine

slashmo: Good idea, I'll look into that too.

adam-fowler:
This list looks fairly good. You did forget find a solution for pipelined commands.

slashmo:
> You did forget find a solution for pipelined commands.

We also briefly discussed this, forgot to mention it here. What I'd ideally like to do is to pass whether the pipelined execute was preceded by a `MULTI` to the variadic generics overload of `execute` so that it can either create a span called `MULTI` or a span called something like `pipeline` with a span attribute. Valkey Swift's transactional API would pass `true` to the `execute` method but by default it'd be `false`. In either case, the span created in `execute` should have a span attribute with the list of pipelined command names.

Tests/ValkeyTests/Utils/TestTracer.swift
ktoso: Yeah lots of folks wrote such tracers, I think we're ready to accept such thing into upstream into a separate target. We can keep the tasklocal lookup separate though;

There can be a tracer that does the recording and inspecting; and the looking up we can leave to how you bootstrap or pass the tracer perhaps

ktoso:
Great to see this happening, thank you @slashmo, @adam-fowler, @fabianfett! Was just discussing otel plans with Moritz and the thread came up.

I'm really glad to see this approach does: 
- depend on swift-distributed-tracing directly, that's _the way_ we should be doing this
- guards the feature with a package trait
- is _on_ by default; I think that's the right choice, otherwise running around and trying to enable all the possible traits would be tricky.

It's somewhat concerning to adopt a pattern of "pass `(any Tracer)?`" because that's somewhat incompatible with any of the ways to use `tracer.withSpan { ... }`. You can't really:

```swift
// bad, always need to run the nested code
tracer?.withSpan(...) { span in code } // ofc wrong, always need to run the 'code'

// so what... duplication like this is absolutely not ok
if tracer {  
  tracer.withSpan(...) { span in code } 
} else {
  code 
}
```

While with noop-tracers this "just works" because: 

```swift
tracer.withSpan(...) { noopSpan in code } // noopspan is a struct
```

Are we absolutely certain this optional-nil vs "noop" is a _must_ here? Is the difference _actually_ significant enough to warrant the pain, given this is in _tracing is enabled_ mode? If so, we must answer the question what libraries should be doing when using tracer like this.

Because what would the pattern look like for adopter code of tracing that _do_ use structured concurrency. I understand your specific lib is not using structured concurrency in the core, but if we start recommending "pass `(any Tracer)?`" I'm concerned this will become a pattern everywhere, and just _not work at all_ for libs doing structured concurrency


We have a weekly call we could purpose for designing the answer here, @slashmo will try to organize the invites! :-)

ktoso (approved):
Looks good! I think that's a nice approach, maybe just add some comments about it

Sources/Valkey/Connection/ValkeyConnection.swift
```diff
@@ -159,16 +181,47 @@ public final actor ValkeyConnection: ValkeyClientProtocol, Sendable {
...
+                    self.channelHandler.write(command: command, continuation: continuation, requestID: requestID)
+                }
+            } onCancel: {
+                self.cancel(requestID: requestID)
             }
-            return try await withCheckedThrowingContinuation { continuation in
-                self.channelHandler.write(command: command, continuation: continuation, requestID: requestID)
+        } catch let error as ValkeyClientError {
+            #if DistributedTracingSupport
+            span?.recordError(error)
```
adam-fowler, line 208: Maybe wrap all this code in `if let span`

Sources/Valkey/Connection/ValkeyConnectionConfiguration.swift
```diff
@@ -142,3 +149,26 @@ public struct ValkeyConnectionConfiguration: Sendable {
         self.clientName = clientName
     }
 }
+
+#if DistributedTracingSupport
+public struct ValkeyTracingConfiguration: Sendable {
+
+    public var tracer: (any Tracer)? = InstrumentationSystem.tracer
```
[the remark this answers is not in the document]
adam-fowler: If global tracer is NoOp tracer is it worth setting `tracer` to `nil`?

Sources/Valkey/Connection/ValkeyConnection.swift
```diff
@@ -161,16 +176,48 @@ public final actor ValkeyConnection: ValkeyClientProtocol, Sendable {
...
-                self.channelHandler.write(command: command, continuation: continuation, requestID: requestID)
+        } catch let error as ValkeyClientError {
+            #if DistributedTracingSupport
+            span.recordError(error)
+            if let message = error.message {
+                var prefixEndIndex = message.startIndex
+                while prefixEndIndex < message.endIndex, message[prefixEndIndex] != " " {
+                    message.formIndex(after: &prefixEndIndex)
+                }
+                let prefix = message[message.startIndex..<prefixEndIndex]
```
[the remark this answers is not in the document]
adam-fowler: Maybe break the code into a separate function as well, to make it clear what you are doing

adam-fowler (commented):
In general I think we are good, added a few more comments.

Sources/Valkey/Connection/ValkeyConnection.swift
```diff
@@ -56,6 +67,17 @@ public final actor ValkeyConnection: ValkeyClientProtocol, Sendable {
         self.configuration = configuration
         self.id = connectionID
         self.logger = logger
+        #if DistributedTracingSupport
+        self.tracer = configuration.tracing.tracer
+        #endif
+        switch address?.value {
```
adam-fowler, line 73: Given `address` is now declared inside a `#if DistributedTracingSupport` its initialisation will also need to be conditional compiled.

Sources/Valkey/Connection/ValkeyConnection.swift
```diff
@@ -219,6 +269,16 @@ public final actor ValkeyConnection: ValkeyClientProtocol, Sendable {
         }
     }
 
+    @usableFromInline
+    func applyCommonAttributes(to attributes: inout SpanAttributes, commandName: String) {
+        attributes[self.configuration.tracing.attributeNames.databaseOperationName] = commandName
```
adam-fowler, line 274: Add `#if DistributedTracingSupport`

Benchmarks/ValkeyBenchmarks/ValkeyConnectionBenchmark.swift
```diff
@@ -60,9 +67,45 @@ func makeConnectionGETBenchmark() -> Benchmark? {
     return Benchmark("Connection: GET benchmark", configuration: .init(metrics: defaultMetrics, scalingFactor: .kilo)) { benchmark in
         let port = serverMutex.withLock { $0 }!.localAddress!.port!
         let logger = Logger(label: "test")
+        #if DistributedTracingSupport
+        // explicitly set tracer to nil, if trait is enabled
+        var configuration = ValkeyConnectionConfiguration()
+        configuration.tracing.tracer = nil
+        #endif
```
adam-fowler, line 74:
```suggestion
        #else
        let configuration = ValkeyConnectionConfiguration()
        #endif
```

adam-fowler (commented):
You need to ensure everything runs with the DistributedTracingSupport trait disabled
```
swift test --disable-default-traits
```

Sources/Valkey/Connection/ValkeyConnectionConfiguration.swift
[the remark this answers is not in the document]
adam-fowler: @slashmo Can we get some comments on ValkeyTracingConfiguration and its members. This is currently a public symbol without any comments.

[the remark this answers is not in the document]
slashmo: Yes, let me add some.

Sources/Valkey/Connection/ValkeyConnection.swift
```diff
@@ -153,16 +175,44 @@ public final actor ValkeyConnection: ValkeyClientProtocol, Sendable {
...
+                    span.attributes["db.response.status_code"] = "\(prefix)"
+                }
             }
-        } onCancel: {
-            self.cancel(requestID: requestID)
+            #endif
+            throw error
+        } catch {
+            #if DistributedTracingSupport
+            span?.recordError(error)
```
czechboy0, line 213: Is it intentional that this code path doesn't set the status like the one above?

PR #178, opened 2025-08-11
Breaking out running client code from the ValkeyClusterClient StateMachine

adam-fowler:
So this can be used by ValkeyClient to manage replica connection pools

### DIFF
--- a/Sources/Valkey/Node/ValkeyRunningClients.swift
+++ b/Sources/Valkey/Node/ValkeyRunningClients.swift
@@ -29,7 +29,7 @@ struct ValkeyRunningClients<
     let poolFactory: ConnectionPoolFactory
     @usableFromInline
     var clientMap: [ValkeyNodeID: NodeBundle]
-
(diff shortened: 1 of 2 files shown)


Sources/Valkey/Node/ValkeyRunningClients.swift
```diff
@@ -0,0 +1,111 @@
...
+//
+// See LICENSE.txt for license information
+// See valkey-swift/CONTRIBUTORS.txt for the list of valkey-swift authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+@usableFromInline
+struct ValkeyRunningClients<
```
fabianfett, line 16: Can we add a `StateMachine` at the end please? Otherwise this sounds very ref typey. Approving anyway.

fabianfett (approved):
Would love to see the one comment addressed.

Issue #184, opened 2025-08-20
Bug: LMOVE does not properly parse "null bulk string" output


aim2120:
## Bug

When using a `ValkeyClusterClient`, I observe when a `LMOVE` command is executed on an empty source array, the client does not properly parse the output `_` (null bulk string). This results in the `ValkeyClusterClient.lmove` function throwing a `RESPParsingError` due to the unexpected output.

## Expected Behavior

The `ValkeyClusterClient.lmove` function should return `nil` if the source array is empty. This aligns with the [Valkey docs for LMOVE](https://valkey.io/commands/lmove/)

> If source does not exist, the value nil is returned and no operation is performed.

## Current Workaround

In my own code, I am simply catching the `RESPParsingError` and ignoring it if the `buffer` (stripped of whitespace) is equal to `_`.

## Environment

valkey-swift version: [0.1.0](https://github.com/valkey-io/valkey-swift/releases/tag/0.1.0)
Swift version: 6.1.2
Xcode version: 16.4
macOS version: 15.6

PR #185, opened 2025-08-20
Fix LMOVE empty source array bug

aim2120:
Fixes the issue https://github.com/valkey-io/valkey-swift/issues/184

Note: for backwards compatibility, I have not changed the `lmove` API. However, this PR does change the `Response` type of the `LMOVE` struct, which is a breaking change.

I am hoping to hear from the maintainers as to how flexible you all are with breaking changes, given the very early age of the package (also it being a version 0 package). If you all are open to it, it may be desirable to change the `lmove` API to be optional in this PR.

Please let me know what you think. I would love to be able to remove the workarounds I have in my current code for this.

### DIFF
--- a/Sources/Valkey/Commands/ListCommands.swift
+++ b/Sources/Valkey/Commands/ListCommands.swift
@@ -303,7 +303,7 @@ public struct LMOVE: ValkeyCommand {
             }
         }
     }
-    public typealias Response = ByteBuffer
+    public typealias Response = ByteBuffer?
 
     public var source: ValkeyKey
     public var destination: ValkeyKey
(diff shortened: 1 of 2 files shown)


adam-fowler:
Before 1.0 we are open to breaking changes.
The command files are generated code. Any changes to them will get lost. 
Frustratingly the docs might mention the `nil` value, but the [model](https://github.com/valkey-io/valkey/blob/unstable/src/commands/lmove.json) file we base the code generation off doesn't. You can see the result of this in the final line of the docs where it says it only returns a bulkString.

We can get around this though. 
- Add "LMOVE" to the list of commands that we don't attempt to generate a response for [here](https://github.com/valkey-io/valkey-swift/blob/600eafad20fbd42f77fc7cb5f8001fe0917288ea/Sources/_ValkeyCommandsBuilder/ValkeyCommandsRender.swift#L17).
- Run `./dev/generate-commands.sh`
- Then add to https://github.com/valkey-io/valkey-swift/blob/main/Sources/Valkey/Commands/Custom/ListCustomCommands.swift
```swift
extension LMOVE {
    public typealias Response = ByteBuffer?
}
```

adam-fowler:
> Should we report this upstream?

Already done. https://github.com/valkey-io/valkey/issues/2532

PR #186, opened 2025-08-21
Add static name to Valkey commands

slashmo:
## Motivation

Most `ValkeyCommand` types in valley-swift exactly match their underlying Valkey command name. However, for some commands like nested ones the Swift type name differs from the underlying command. This means there's no simple way of accessing the underlying command's name such as using `"\(type(of: Self.self))"`, which is required for Distributed Tracing support to derive proper span operation names based on the command being executed. See Also: https://github.com/valkey-io/valkey-swift/pull/177
When running codegen for these commands we _do_ have access to the underlying command names, so we could add them as properties to the generated Swift types.

> [!Note]
> This PR is split off from #177 to decrease that PR's diff and keep it focused on just adding Distributed Tracing support.

## Modifications

- Updated the command builder to generate a command name property, to be used in Distributed Tracing spans
- Re-run command builder to generate command name properties

## Result

All `ValkeyCommand` conforming types now have a static `name` property exposing the underlying Valkey command name. This will enable proper Distributed Tracing span operation names.

### DIFF
--- a/Sources/Valkey/Commands/BitmapCommands.swift
+++ b/Sources/Valkey/Commands/BitmapCommands.swift
@@ -82,6 +82,8 @@ public struct BITCOUNT: ValkeyCommand {
     }
     public typealias Response = Int
 
+    @inlinable public static var name: String { "BITCOUNT" }
+
     public var key: ValkeyKey
     public var range: Range?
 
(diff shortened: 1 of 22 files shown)


slashmo:
> You forgot to include the ValkeyCommandsBuilder changes

Those changes were already in the PR (https://github.com/valkey-io/valkey-swift/pull/186/files#diff-2f98614f8e9a53691e3637ff419218f0aaf7554d5d9eb48f6572c3087388e002). I did however forget to add the `name` property to the non-generated `Test` command in `IntegrationTests`.

PR #187, opened 2025-08-24
Valkey 9.0.0

adam-fowler:
This adds the command changes for valkey 9.0
I had to increase the number of possible parameters for the multi command encoder functions to 29 because `CLIENT LIST` has an additional 16 parameters on top of the 13 it already has.

### DIFF
--- a/Sources/Valkey/Commands/GeoCommands.swift
+++ b/Sources/Valkey/Commands/GeoCommands.swift
@@ -41,11 +41,12 @@ public struct GEOADD<Member: RESPStringRenderable>: ValkeyCommand {
         }
     }
     public struct Data: RESPRenderable, Sendable, Hashable {
-        @usableFromInline let longitude: Double
-        @usableFromInline let latitude: Double
-        @usableFromInline let member: Member
+        public var longitude: Double
+        public var latitude: Double
+        public var member: Member
 
-        @inlinable public init(longitude: Double, latitude: Double, member: Member) {
+        @inlinable
+        public init(longitude: Double, latitude: Double, member: Member) {
             self.longitude = longitude
             self.latitude = latitude
             self.member = member
@@ -200,10 +201,11 @@ public struct GEORADIUS: ValkeyCommand {
         }
     }
     public struct CountBlock: RESPRenderable, Sendable, Hashable {
-        @usableFromInline let count: Int
-        @usableFromInline let any: Bool
+        public var count: Int
+        public var any: Bool
 
-        @inlinable public init(count: Int, any: Bool = false) {
+        @inlinable
+        public init(count: Int, any: Bool = false) {
             self.count = count
             self.any = any
         }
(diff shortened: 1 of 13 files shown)


dev/generate-multi-command-encoder.sh
```diff
@@ -53,7 +54,7 @@ echo "extension ValkeyCommandEncoder {"
 # note:
 # - widening the interval below (eg. going from {1..15} to {1..25}) is Semver minor
 # - narrowing the interval below is SemVer _MAJOR_!
-for n in {1..15}; do
```
fabianfett, line 56: I think, that we should do all the versions above 10 with a parameter pack implementation. But that can be a follow up.

Sources/Valkey/Commands/ServerCommands.swift
```diff
@@ -1444,23 +1426,36 @@ public struct SHUTDOWN: ValkeyCommand {
...
+        @usableFromInline let safe: Bool
+        @usableFromInline let failover: Bool
+
+        @inlinable public init(
+            saveSelector: AbortSelectorSaveSelectorBlockSaveSelector? = nil,
+            now: Bool = false,
+            force: Bool = false,
+            safe: Bool = false,
+            failover: Bool = false
+        ) {
```
fabianfett, line 1438: We will need to come up with a plan so that this isn't an API break anymore, once we have tagged 1.0

Sources/Valkey/Commands/ServerCommands.swift
```diff
@@ -478,25 +472,19 @@ public enum COMMANDLOG {
     @_documentation(visibility: internal)
     public struct LEN: ValkeyCommand {
         public enum _Type: RESPRenderable, Sendable, Hashable {
```
fabianfett, line 474: we will need to add code gen tricks so that this isn't an API break anymore. Or we just support 6.2 and above with https://github.com/swiftlang/swift-evolution/blob/main/proposals/0487-extensible-enums.md

Sources/Valkey/Commands/GeoCommands.swift
```diff
@@ -1030,15 +1073,57 @@ public struct GEOSEARCHSTORE: ValkeyCommand {
...
+            self.vertices = vertices
+        }
+
+        @inlinable
+        public var respEntries: Int {
+            RESPWithToken("BYPOLYGON", numVertices).respEntries + vertices.respEntries
+        }
+
+        @inlinable
+        public func encode(into commandEncoder: inout ValkeyCommandEncoder) {
```
fabianfett, line 1111: Can we make the placement of `@inlinable`/`@usableFromInline` consistent? I like the above the function call/variable definition version most. Follow up PR is fine.

Sources/Valkey/Commands/GeoCommands.swift
```diff
@@ -1030,15 +1073,57 @@ public struct GEOSEARCHSTORE: ValkeyCommand {
             unit.encode(into: &commandEncoder)
         }
     }
+    public struct ByPolygonVertices: RESPRenderable, Sendable, Hashable {
+        @usableFromInline let longitude: Double
+        @usableFromInline let latitude: Double
```
fabianfett, line 1078: I think we should make the properties public and var. This will also help with API evolution.

dev/generate-multi-command-encoder.sh
adam-fowler: But the parameter pack version crashes the release build, otherwise we would use a parameter pack all the time

PR #191, opened 2025-08-25
Add subscription connection manager

adam-fowler:
Manages creation of single connection for subscriptions


Sources/Valkey/Subscriptions/SubscriptionConnectionManager.swift
```diff
@@ -0,0 +1,140 @@
...
+    @usableFromInline
+    let state: Mutex<State>
+
+    init() {
+        self.state = .init(.uninitialized)
+    }
+
+    @usableFromInline
+    package func acquire(
+        isolation: isolated (any Actor)? = #isolation,
```
fabianfett, line 46: I think we should run the `acquire` action inside the client's background task, otherwise we run into issues with cancellation again.

PR #209, opened 2025-09-06
Cluster command keys should all come from the same hash slot

adam-fowler:
This allows us to simplify a bunch of internal code as we only need to know the shard for a single hash key


adam-fowler:
Reducing this PR just to the changes in ValkeyClusterClient. We can fix the underlying APIs in a separate PR

nilanshu-sharma (approved):
LGTM!

PR #219, opened 2025-09-18
Cluster subscribe

adam-fowler:
ValkeyClient has a subscribe command that allocates a single connection for all subscriptions. This PR adds something similar except it allocates a single connection from each primary node in the cluster (once we have reading from replicas we can extend to replicas).

To get this working 
- I had to move the subscription connection management code from ValkeyClient to ValkeyNodeClient. 
- I added an extra state `shutdown` to the subscription connection state machine.
- I added a `shutdownSubscriptionConnection` function that sets the shutdown subscription and resumes the release connection continuation.
- When subscribe is called cluster client chooses random node and uses subscription connection from that node.


Sources/Valkey/Node/ValkeyNodeClient.swift
```diff
@@ -47,6 +54,17 @@ package final class ValkeyNodeClient: Sendable {
     public let eventLoopGroup: any EventLoopGroup
     /// Logger
     public let logger: Logger
+    /// subscription connection state
+    @usableFromInline
+    let subscriptionConnectionStateMachine: Mutex<ConnectionStateMachine>
```
nilanshu-sharma, line 59: Minor comment: Should be this called `subscriptionConnectionStateMachineMutex` or `subscriptionConnectionStateMachineLock` as it is a mutex around  `ConnectionStateMachine`

adam-fowler: I'm happy with its name. It is a bit more than a mutex as it manages the leasing of the connection as well

Sources/Valkey/Subscriptions/ValkeyClient+subscribe.swift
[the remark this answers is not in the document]
adam-fowler: The `+` indicates the code in this file is an extension. In this case it is `ValkeyClient` being extended to include subscription functionality

PR #222, opened 2025-09-18
Add pipeline functions that take an array of existential ValkeyCommands

adam-fowler:
This allows us to dynamically create pipelines, in a way we can't with parameter packs. In theory it should run slightly slower than the parameter pack version, in that you have to create an array of commands, the commands are stored as existential and an array for the results has to be created. I added a benchmark and the performance difference isn't that big.

Also ValkeyClient functions now call directly into the ValkeyNodeClient versions of the pipeline functions

### DIFF
--- a/Sources/Valkey/Connection/ValkeyConnection.swift
+++ b/Sources/Valkey/Connection/ValkeyConnection.swift
@@ -277,15 +277,6 @@ public final actor ValkeyConnection: ValkeyClientProtocol, Sendable {
     public func execute(
         _ commands: some Collection<any ValkeyCommand>
     ) async -> sending [Result<RESPToken, Error>] {
-        func convert<Response: RESPTokenDecodable>(_ result: Result<RESPToken, Error>, to: Response.Type) -> Result<Response, Error> {
-            result.flatMap {
-                do {
-                    return try .success(Response(fromRESP: $0))
-                } catch {
-                    return .failure(error)
-                }
-            }
-        }
         let requestID = Self.requestIDGenerator.next()
         // this currently allocates a promise for every command. We could collapse this down to one promise
         var mpromises: [EventLoopPromise<RESPToken>] = []


Sources/Valkey/Connection/ValkeyConnection.swift
```diff
@@ -261,6 +261,63 @@ public final actor ValkeyConnection: ValkeyClientProtocol, Sendable {
...
+    /// is more expensive to run and the command responses are returned as ``RESPToken``
+    /// instead of the response type for the command.
+    ///
+    /// - Parameter commands: Collection of ValkeyCommands
+    /// - Returns: Array holding the RESPToken responses of all the commands
+    @inlinable
+    public func execute(
+        _ commands: some Collection<any ValkeyCommand>
+    ) async -> sending [Result<RESPToken, Error>] {
+        func convert<Response: RESPTokenDecodable>(_ result: Result<RESPToken, Error>, to: Response.Type) -> Result<Response, Error> {
```
nilanshu-sharma, line 280: I don't see we are using this inner function in the current PR. Is this meant for use later ?

Sources/Valkey/Node/ValkeyNodeClient.swift
```diff
@@ -156,6 +156,31 @@ extension ValkeyNodeClient {
...
+    @inlinable
+    public func execute<Commands: Collection & Sendable>(
+        _ commands: Commands
+    ) async -> sending [Result<RESPToken, Error>] where Commands.Element == any ValkeyCommand {
+        do {
+            return try await self.withConnection { connection in
+                await connection.execute(commands)
+            }
+        } catch {
+            return .init(repeating: .failure(error), count: commands.count)
```
nilanshu-sharma, line 181: Do we need to call convert() on the pipeline responses so that errors have `.failure()` ?

nilanshu-sharma (changes requested):
Looks good overall, just seeking some clarification error conversions.

Sources/Valkey/Node/ValkeyNodeClient.swift
nilanshu-sharma: Good point. We too have faced this problem in the past. So we have 2 different pipeline functions:
1. sync() - which doesn't throw an exception. It is like a best effort pipeline sync 
2. sync With Failure() - which inspects the pipeline buffer for responses and throws exception if that failed. 
We can add functions with similar semantics later.

nilanshu-sharma (approved):

PR #263, opened 2025-11-26
Add ValkeySearch commands

eric-musliner:
This is dependent on the work to update the command json files in the ValkeySearch project https://github.com/valkey-io/valkey-search/pull/504

### DIFF
--- a/Sources/ValkeySearch/SearchCommands.swift
+++ b/Sources/ValkeySearch/SearchCommands.swift
@@ -378,23 +378,45 @@ public enum FT {
                 milliseconds.encode(into: &commandEncoder)
             }
         }
-        public struct Params: RESPRenderable, Sendable, Hashable {
-            public var nargs: Int
-            public var parameters: [String]
+        public struct ParamsParameters: RESPRenderable, Sendable, Hashable {
+            public var name: String
+            public var value: String
 
             @inlinable
-            public init(nargs: Int, parameters: [String]) {
+            public init(name: String, value: String) {
+                self.name = name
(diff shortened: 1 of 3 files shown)


eric-musliner:
> It's great you've added the updates to the spec files. I believe this'll also improve the documentation. I can't really merge this PR until the spec files PR is merged (and ideally in a release). We are near releasing a 1.0 and that will mean breaking changes can't be merged. We can workaround changes to the API caused by changes to the json spec files, but I don't really want to be doing this multiple times. We could hide these behind an "experimental-json-search" trait which we document as an API that may change.

Completely understand. This PR is dependent on the JSON changes getting approved and merged in ValkeySearch. I definitely want eyes on that one to make sure it's correct and that I'm not missing anything.

eric-musliner:
> I'm wondering if we can release this behind a `ExperimentalValkeySearch` trait. I've been bitten using traits with Hummingbird so uncertain about this.

I've never used traits before, but the valkey search commands don't break the existing APIs so even if you do put out a 1.0.0 release before this gets merged in you could include the valkey search commands in a minor version update? Unless you're rationale is to bring it in as a trait since now since we are still waiting on confirmation about json docs?

eric-musliner:
> The rationale of adding it behind a trait tagged experimental is we are not committing to the public API. So if there is an error in the generated code we don't have to worry about breaking changes.

That makes sense to me if you want to bring it in before the json documents are reviewed and merged.

adam-fowler (approved):
I think we can merge this as is.

PR #269, opened 2025-12-02
Command array arguments that render with a token for each element

adam-fowler:
- Add RESPArrayWithToken that renders each element with token prefix
- Use RESPArrayWithToken for elements with flag `multiple_token`
- Use `CLIENT TRACKING` to test this


philippzagar (approved):
Looks good from my perspective!

Issue #261, opened 2025-11-13
Adopt upcoming nonisolated(nonsending) by default


fabianfett:
- Enable feature flag
- Remove all isolated parameters

fabianfett:
https://github.com/vapor/postgres-nio/pull/604

PR #272, opened 2025-12-04
Adopt `NonisolatedNonsendingByDefault`

fabianfett:
Fixes: #261.

### DIFF
--- a/Sources/Valkey/Subscriptions/ValkeyConnection+subscribe.swift
+++ b/Sources/Valkey/Subscriptions/ValkeyConnection+subscribe.swift
@@ -72,7 +72,6 @@ extension ValkeyConnection {
         }
     }
 
-    #if compiler(>=6.2)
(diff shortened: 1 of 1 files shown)


Sources/Valkey/Subscriptions/ValkeySubscription.swift
```diff
@@ -51,8 +51,19 @@ public struct ValkeySubscription: AsyncSequence, Sendable {
     public struct AsyncIterator: AsyncIteratorProtocol {
         var base: BaseAsyncSequence.AsyncIterator
 
+        #if compiler(>=6.2)
+        @concurrent
```
adam-fowler: Could this be 
```swift
public mutating func next() async throws -> Element? {
    try await self.base.next(isolation: #isolation)
}
```
Instead of adding `@concurrent`

PR #273, opened 2025-12-04
Custom responses list commands

nilanshu-sharma:
Adding custom responses for `lpop` an `rpop` commands which return: 
* `nil` if key doesn't exist 
* `ByteBuffer?` if invoked without `count` param 
* `[ByteBuffer]?` if invoked with `count`

### DIFF
--- a/Tests/ValkeyTests/CommandTests.swift
+++ b/Tests/ValkeyTests/CommandTests.swift
@@ -300,12 +300,19 @@ struct CommandTests {
                 var lpopResponse = try await connection.lpop("key1")
                 #expect(lpopResponse != nil)
                 #expect(try lpopResponse!.element() == ByteBuffer(string: "one"))
-                #expect(try lpopResponse!.elements() == nil)
+                #expect(try lpopResponse!.elements() == [ByteBuffer(string: "one")])
 
                 // Test multiple elements LPOP with count
                 lpopResponse = try await connection.lpop("key1", count: 3)
                 #expect(lpopResponse != nil)
-                #expect(try lpopResponse!.element() == nil)
+                do {
+                    _ = try lpopResponse!.element()
+                    Issue.record("Expected RESPDecodeError.tokenMismatch to be thrown")
+                } catch let error as RESPDecodeError {
+                    #expect(error.errorCode == .tokenMismatch)
+                } catch {
+                    Issue.record("Unexpected error type: \(error)")
+                }
                 #expect(try lpopResponse!.elements() == [ByteBuffer(string: "two"), ByteBuffer(string: "three"), ByteBuffer(string: "four")])
             }
         }


Sources/Valkey/Commands/Custom/ListCustomCommands.swift
```diff
@@ -63,3 +63,46 @@ extension BRPOP {
     ///     * [Array]: The key from which the element was popped and the value of the popped element
     public typealias Response = ListEntry?
 }
+
+/// Custom response type for LPOP and RPOP commands
+/// Handles the different return types based on whether count parameter is provided
+@_documentation(visibility: internal)
+public struct ListPopResponse: RESPTokenDecodable, Sendable, Hashable {
```
adam-fowler: I think we should this in a similar manner to how you did HRANDFIELD where we store a RESPToken internally and decode when the element or elements function are called

nilanshu-sharma: > I think we should this in a similar manner to how you did HRANDFIELD where we store a RESPToken internally and decode when the element or elements function are called

Good point, I have reworked the logic to be similar to `hrandfield`.

Sources/Valkey/Commands/Custom/ListCustomCommands.swift
```diff
@@ -63,3 +63,31 @@ extension BRPOP {
...
+
+    public init(_ token: RESPToken) throws {
+        self.token = token
+    }
+
+    /// Gets the single element when count was not provided
+    /// - Returns: ByteBuffer if a single element was returned, nil otherwise
+    public func element() throws -> ByteBuffer? {
+        // Handle .null as it is expected when the key doesn't exist
+        if token.value == .null {
```
adam-fowler: Checking for null here shouldn't be needed, that has already been done because the LPOP/RPOP response is optional. Which means the function can return a non-optional value. It should be `RESPBulkString` instead of `ByteBuffer`

Sources/Valkey/Commands/Custom/ListCustomCommands.swift
```diff
@@ -63,3 +63,31 @@ extension BRPOP {
...
+            return nil
+        }
+        return try ByteBuffer(token)
+    }
+
+    /// Gets the multiple elements when count was provided
+    /// - Returns: Array of ByteBuffer if multiple elements were returned, nil otherwise
+    public func elements() throws -> [ByteBuffer]? {
+        try [ByteBuffer]?(token)
```
adam-fowler: This doesn't need to be optional

adam-fowler (commented):
Some additional comments.

This PR does make me wonder about the sense of creating responses for every command instead of allowing some just to return a RESPToken. Note the amount of extra code you needed to add in the `CommandTests` test

PR #284, opened 2025-12-18
Add tracing spans for pipeline and transaction functions

adam-fowler:
This add tracing spans for the pipeline and transaction functions. Span name is either `Pipeline` or `MULTI`. The database operation name though is a list of the commands concatenated together eg `SET,SET,INCR`.

None of the functions report errors for individual commands as the Results returned by the commands are not processed and processing them would add additional  The pipeline functions don't report errors as it doesn't process errors it only returns a collection of Results. The transaction functions report errors returned by `EXEC` but again errors from individual commands are not reported.

@slashmo


Sources/Valkey/Connection/ValkeyConnection.swift
```diff
@@ -174,7 +175,8 @@ public final actor ValkeyConnection: ValkeyClientProtocol, Sendable {
         defer { span?.end() }
 
         span?.updateAttributes { attributes in
-            self.applyCommonAttributes(to: &attributes, commandName: Command.name)
+            self.applyCommonAttributes(to: &attributes)
+            attributes[self.configuration.tracing.attributeNames.databaseOperationName] = Command.name
```
nilanshu-sharma: Should we also add databaseNumber here as well for better tracing ?

Sources/Valkey/Connection/ValkeyConnection.swift
```diff
@@ -587,35 +664,51 @@ public final actor ValkeyConnection: ValkeyClientProtocol, Sendable {
...
+            stringLength += Swift.type(of: command).name.count + 1
+        }
         var string: String = ""
+        string.reserveCapacity(stringLength - 1)
+
         var count = 0
         for command in repeat each commands {
             if count == 0 {
                 string += "\(Swift.type(of: command).name)"
-            } else if count == 16 {
```
nilanshu-sharma: Any reason we are removing this logic of having at most 16 commands here but we have the same logic in the `concatenateCommandNames` function below ? 
I feel we can just cap the entries to size 16 and keep the behavior consistent.

Sources/Valkey/Connection/ValkeyConnection.swift
```diff
@@ -320,11 +361,24 @@ public final actor ValkeyConnection: ValkeyClientProtocol, Sendable {
...
+    public func execute<Commands: Collection & Sendable>(
+        _ commands: Commands
+    ) async -> [Result<RESPToken, any Error>] where Commands.Element == any ValkeyCommand {
         self.logger.trace("execute", metadata: ["commands": .string(Self.concatenateCommandNames(commands))])
 
+        #if DistributedTracingSupport
+        let span = self.tracer?.startSpan("Pipeline", ofKind: .client)
+        defer { span?.end() }
+
+        if !(span is NoOpTracer.Span) {
```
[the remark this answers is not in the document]
adam-fowler: How the attributes are updated is different for each occurrence.

PR #287, opened 2025-12-29
ValkeyClient state machine. Sets up primary node and will search for replicas if required

adam-fowler:
This PR does not include the code for using the replicas yet. That will be in a separate PR


nilanshu-sharma (approved):
LGTM

PR #290, opened 2026-01-06
Cluster nodes custom response

nilanshu-sharma:
Adding custom reponses for following commands: 

* `CLUSTER NODES`
* `CLUSTER REPLICAS`
* `CLUSTER SLAVES` 

All 3 commands return the same response. This is to prevent repeated decoding of the responses by users. 
Added integration tests to verify the logic. 

After this change all `CLUSTER *` commands should have custom responses.

### DIFF
--- a/Tests/ClusterIntegrationTests/ClusterIntegrationTests.swift
+++ b/Tests/ClusterIntegrationTests/ClusterIntegrationTests.swift
@@ -696,7 +696,7 @@ struct ClusterIntegrationTests {
                 [(host: firstNodeHostname, port: firstNodePort)],
                 logger: logger
             ) { client in
-                let nodeSelections: [ValkeyClusterNodeSelection] = [.primary, .cycleReplicas(234)]
+                let nodeSelections: [ValkeyNodeSelection] = [.primary, .cycleReplicas(234)]
                 for selection in nodeSelections {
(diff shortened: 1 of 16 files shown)


Tests/ClusterIntegrationTests/ClusterIntegrationTests.swift
```diff
@@ -433,6 +433,54 @@ struct ClusterIntegrationTests {
         }
     }
 
+    @Test
```
adam-fowler: Can we do the tests in a similar manner to `CommandTests`. We shouldn't need a live server to test these

Sources/Valkey/Commands/Custom/ClusterCustomCommands.swift
```diff
@@ -35,6 +35,175 @@ extension CLUSTER.SLOTS {
...
+    /// Last ping sent timestamp
+    public var pingSent: Int64
+    /// Last pong received timestamp
+    public var pongReceived: Int64
+    /// Configuration epoch for this node
+    public var configEpoch: Int64
+    /// Link state to this node (connected or disconnected)
+    public var linkState: String
+    /// Hash slots served by this node (only for primaries)
+    public var slots: [String]
```
adam-fowler: Would be nice if these slots were parsed. And provided the ranges instead of a string

adam-fowler: fair enough

PR #302, opened 2026-01-12
MEMORY STATS custom command

adam-fowler:
MEMORY STATS has so many values, I decided to just return a map for this. I have created a key type so users can find the values easier. They will have to convert them to the correct type though.

```swift
let stats = try await client.memoryStats()
print(stats[.totalAllocated].decode(as: Int.self))
```

### DIFF
--- a/Sources/Valkey/Commands/Custom/ServerCustomCommands.swift
+++ b/Sources/Valkey/Commands/Custom/ServerCustomCommands.swift
@@ -226,6 +226,90 @@ extension ROLE {
     }
 }
 
+extension MEMORY.STATS {
+    public struct Key: RawRepresentable, RESPTokenDecodable, Hashable, Sendable, CustomStringConvertible {
+        public let rawValue: String
+
+        public init(rawValue: String) {
+            self.rawValue = rawValue
+        }
+
+        public init(_ token: RESPToken) throws {
+            let string = try String(token)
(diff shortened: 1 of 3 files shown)


Sources/Valkey/Commands/Custom/ServerCustomCommands.swift
```diff
@@ -108,3 +108,87 @@ extension ROLE {
...
+        public static var rssOverheadRatio: Key { .init(rawValue: "rss-overhead.ratio") }
+        /// See INFO's rss_overhead_bytes
+        public static var rssOverheadBytes: Key { .init(rawValue: "rss-overhead.bytes") }
+        /// See INFO's mem_fragmentation_ratio
+        public static var fragmentation: Key { .init(rawValue: "fragmentation") }
+        /// See INFO's mem_fragmentation_bytes
+        public static var fragmentationBytes: Key { .init(rawValue: "fragmentation.bytes") }
+    }
+
+    public typealias Response = [Key: RESPToken]
```
nilanshu-sharma: Quick Question: Does this declaration automatically convert the `RESPToken` to `Response` ?

adam-fowler: Yes this is the type that the Response is converted to. So in this case `memoryStats` returns a `[Key: RESPToken]`

PR #307, opened 2026-01-13
ZRANDMEMBER Custom response

nilanshu-sharma:


Sources/Valkey/Commands/Custom/SortedSetCustomCommands.swift
```diff
@@ -120,3 +120,58 @@ extension ZSCAN {
...
+
+        /// Get multiple random member-score pairs when ZRANDMEMBER was called with COUNT and WITHSCORES
+        /// - Returns: Array of SortedSetEntry (member-score pairs), or nil if key doesn't exist
+        /// - Throws: RESPDecodeError if response format is unexpected
+        public func multipleMembersWithScores() throws -> [SortedSetEntry]? {
+            switch token.value {
+            case .null:
+                return nil
+            case .array(let array):
+                var iterator = array.makeIterator()
```
adam-fowler: You can shorten the extracting of the member and score code by converting to a RESPToken.Map 
```swift
var array: [SortedSetEntry] = []
for respElement in try self.elements.asMap() {
    let value = try RESPBulkString(respElement.key)
    let score = try Double(respElement.value)
    array.append(.init(value: value, score: score))
}
return array
```
If you were being really clever you could also do the following
```swift
return try self.elements.asMap().map {
    try SortedSetEntry(value: RESPBulkString($0.key), score: Double($0.value))
}
```

PR #311, opened 2026-01-15
Added typed throws to RESPTokenDecodable(_:)

adam-fowler:


Sources/Valkey/Commands/Custom/ServerCustomCommands.swift
```diff
@@ -213,6 +213,10 @@ extension ROLE {
                     }
                 } catch let error as MissingValueDecodeError {
```
nilanshu-sharma: Unrelated to this PR but we can get rid of MissingValueDecodeError completely and just throw `RESPDecodeError.invalidArraySize`

Sources/Valkey/Commands/Custom/ServerCustomCommands.swift
```diff
@@ -213,6 +213,10 @@ extension ROLE {
...
                     throw RESPDecodeError.invalidArraySize(array, expectedSize: error.expectedNumberOfValues + 1)
+                } catch let error as RESPDecodeError {
+                    throw error
+                } catch {
+                    preconditionFailure("Shouldn't get here")
```
nilanshu-sharma: From what I am reading, this will cause a crash. Should we throw an error instead ? 
I'm wondering if failing to decode a RESP token should crash an app.

Sources/Valkey/Commands/Custom/ServerCustomCommands.swift
nilanshu-sharma: I am happy to do this after this PR is merged :)

nilanshu-sharma (approved):
LGTM!

PR #313, opened 2026-01-19
CLIENT LIST Decoded Response

nilanshu-sharma:
- Adding Custom Response for `CLIENT LIST` command 
- Added unit test to verify the functionality
- Disabled `Response:` comment header generation for custom responses 
- Copied `SplitStringSequences.swift‎` from [Hummingbird](https://github.com/hummingbird-project/hummingbird/blob/main/Sources/Hummingbird/Utils/SplitStringSequences.swift)

### DIFF
--- a/Sources/Valkey/Commands/ConnectionCommands.swift
+++ b/Sources/Valkey/Commands/ConnectionCommands.swift
@@ -1028,7 +1028,7 @@ extension ValkeyClientProtocol {
     ///     * 6.0.0: Added ACL style (username and password).
     /// - Complexity: O(N) where N is the number of passwords defined for the user
     @inlinable
-    public func auth<Password: RESPStringRenderable>(username: String? = nil, password: Password) async throws {
+    public func auth<Password: RESPStringRenderable>(username: String? = nil, password: Password) async throws(ValkeyClientError) {
         _ = try await execute(AUTH(username: username, password: password))
     }
 
@@ -1039,7 +1039,7 @@ extension ValkeyClientProtocol {
     /// - Complexity: Depends on subcommand.
     @inlinable
     @discardableResult
-    public func client() async throws -> CLIENT.Response {
+    public func client() async throws(ValkeyClientError) -> CLIENT.Response {
         try await execute(CLIENT())
     }
 
@@ -1049,7 +1049,7 @@ extension ValkeyClientProtocol {
     /// - Available: 8.0.0
     /// - Complexity: O(1)
     @inlinable
-    public func clientCapa<Capability: RESPStringRenderable>(capabilities: [Capability]) async throws {
+    public func clientCapa<Capability: RESPStringRenderable>(capabilities: [Capability]) async throws(ValkeyClientError) {
         _ = try await execute(CLIENT.CAPA(capabilities: capabilities))
     }
 
@@ -1064,7 +1064,7 @@ extension ValkeyClientProtocol {
     ///     * [Integer]: ID of the client we are redirecting the notifications to.
     @inlinable
     @discardableResult
-    public func clientGetredir() async throws -> Int {
+    public func clientGetredir() async throws(ValkeyClientError) -> Int {
         try await execute(CLIENT.GETREDIR())
     }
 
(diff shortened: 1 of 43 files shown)


Sources/Valkey/Commands/ConnectionCommands.swift
```diff
@@ -1131,7 +1129,7 @@ extension ValkeyClientProtocol {
     ///     * 8.1.0: Added filters USER, ADDR, LADDR, SKIPME, and MAXAGE.
     ///     * 9.0.0: Added filters NAME, IDLE, FLAGS, LIB-NAME, LIB-VER, DB, CAPA, and IP. And negative filters NOT-ID, NOT-TYPE, NOT-ADDR, NOT-LADDR, NOT-USER, NOT-FLAGS, NOT-NAME, NOT-LIB-NAME, NOT-LIB-VER, NOT-DB, NOT-CAPA, NOT-IP.
     /// - Complexity: O(N) where N is the number of client connections
-    /// - Response: [String]: Information and statistics about client connections
```
adam-fowler: This will get overwritten by the code generator

Sources/Valkey/Commands/Custom/ConnectionCustomCommands.swift
```diff
@@ -53,3 +53,148 @@ extension CLIENT.TRACKINGINFO {
...
+
+                let lines = string.split(separator: "\n").map(String.init).filter { !$0.isEmpty }
+
+                var clients: [[Field: RESPToken]] = []
+
+                for line in lines {
+                    var client: [Field: RESPToken] = [:]
+
+                    // Split by spaces and parse key=value pairs
+                    let components = line.split(separator: " ").map(String.init)
```
adam-fowler: You use `split` quite a bit. Every time you use it, it allocates at least once when creating the array to hold the splits. I have this helper class from Hummingbird, which create a sequence of Substrings instead without doing the allocation. Basically every time you ask for the next string it parses the main String. You won't get a count from the sequence but you can iterate the sequence to find out how parts there are. Also don't convert the Substring to a String until it is truly necessary. Here is the code for the split sequence https://github.com/hummingbird-project/hummingbird/blob/main/Sources/Hummingbird/Utils/SplitStringSequences.swift.

Sources/Valkey/Commands/Custom/ConnectionCustomCommands.swift
```diff
@@ -53,3 +53,153 @@ extension CLIENT.TRACKINGINFO {
...
+    }
+
+    /// Response type for CLIENT LIST command.
+    ///
+    /// Returns an array of client information dictionaries, where each dictionary
+    /// maps field names to their RESP token values. This approach gracefully handles
+    /// new fields that may be added in future Valkey versions.
+    public struct Response: RESPTokenDecodable, Sendable {
+        /// Array of client information dictionaries
+        public let clients: [[Field: RESPToken]]
```
adam-fowler: I don't think we need to return this a dictionary of RESPTokens, we are treating the values as Strings internally.

I guess you could even store them as Substrings and set the Field rawValue to be Substring to avoid allocating memory for each field.
```suggestion
        public let clients: [[Field: Substring]]
```

Sources/Valkey/Commands/Custom/ConnectionCustomCommands.swift
```diff
@@ -53,3 +53,153 @@ extension CLIENT.TRACKINGINFO {
...
+            // Use SplitStringSequence for efficient parsing
+            for line in string.splitSequence(separator: "\n") {
+                var client: [Field: RESPToken] = [:]
+
+                // Split by spaces and parse key=value pairs
+                for component in line.splitSequence(separator: " ") {
+                    if !component.contains("=") {
+                        continue
+                    }
+                    let parts = Array(component.splitMaxSplitsSequence(separator: "=", maxSplits: 1))
```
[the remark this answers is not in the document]
adam-fowler: If you store everything as Substrings you can get rid of the String conversion as well

PR #317, opened 2026-01-22
Call READONLY always on cluster nodes (primary and replicas)

adam-fowler:
Call READONLY on all cluster connections regardless of whether we are connecting to a primary or replica. Primaries ignore the READONLY flag and doing this means we don't need to track whether it has been called.

Also extend the devcontainer docker-compose file to have 2 replicas per shard


nilanshu-sharma (approved):
Looks good to me. This will run `READONLY` at the connection init time which enable reads on the replicas, and will be a no-op on the leader. Just a minor comment - now that `readOnly` is always true, we can probably just remove it entirely and run the `READONLY` command unconditionally in ValkeyChannelHandler.setConnected().

PR #318, opened 2026-01-27
Add support for client capa redirect

adam-fowler:
See https://valkey.io/commands/client-capa/
- Add `enableClientCapaRedirect` flag to `ValkeyClientConfiguration`
- If flag is set call `CLIENT CAPA redirect` in channel handler
- When enabled we will receive redirect errors when trying to call a mutating call on a replica
- When we receive a redirect error we retry using the node indicated in the redirect error and set the new node to be the primary.

The PR currently only does redirection on single command execute
This PR also removes the need for the readonly flag for standalone client nodes

### DIFF
--- a/Sources/Valkey/Connection/ValkeyConnectionConfiguration.swift
+++ b/Sources/Valkey/Connection/ValkeyConnectionConfiguration.swift
@@ -129,7 +129,7 @@ public struct ValkeyConnectionConfiguration: Sendable {
     /// Valkey 8.0 introduced a new command CLIENT CAPA to indicate client capabilities. It
     /// currently only supports one capability `redirect`. This indicates the client is
     /// capable of handling redirect messages from replica nodes back to the primary
-    var enableClientRedirect: Bool
+    public var enableClientCapaRedirect: Bool
 
     #if DistributedTracingSupport
     /// The distributed tracing configuration to use for this connection.
@@ -150,6 +150,7 @@ public struct ValkeyConnectionConfiguration: Sendable {
     ///   - clientName: Optional name to identify this client connection on the server. Defaults to `nil`.
     ///   - readOnly: Is the connection a readonly connection
     ///   - databaseNumber: Database Number to use for the connection
+    ///   - enableClientCapaRedirect: Support client redirection errors from replicas
     public init(
         authentication: Authentication? = nil,
         commandTimeout: Duration = .seconds(30),
@@ -158,7 +159,7 @@ public struct ValkeyConnectionConfiguration: Sendable {
         clientName: String? = nil,
         readOnly: Bool = false,
         databaseNumber: Int = 0,
-        enableClientRedirect: Bool = true
+        enableClientCapaRedirect: Bool = true,
     ) {
         self.authentication = authentication
         self.commandTimeout = commandTimeout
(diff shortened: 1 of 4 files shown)


Sources/Valkey/Connection/ValkeyConnectionConfiguration.swift
```diff
@@ -124,6 +124,13 @@ public struct ValkeyConnectionConfiguration: Sendable {
     /// The number of Valkey Database
     public var databaseNumber: Int = 0
 
+    /// Enable client redirect capability
+    ///
+    /// Valkey 8.0 introduced a new command CLIENT CAPA to indicate client capabilities. It
+    /// currently only supports one capability `redirect`. This indicates the client is
+    /// capable of handling redirect messages from replica nodes back to the primary
+    var enableClientRedirect: Bool
```
nilanshu-sharma: Noticing this var is not public unlike other vars in this struct. Perhaps this needs to be public so that it can be exposed to the user for logging etc.

Sources/Valkey/ValkeyClientConfiguration.swift
```diff
@@ -230,6 +230,23 @@ public struct ValkeyClientConfiguration: Sendable {
...
+    /// Enable client redirect capability
+    ///
+    /// Valkey 8.0 introduced a new command CLIENT CAPA to indicate client capabilities. It
+    /// currently only supports one capability `redirect`. This indicates the client is
+    /// capable of handling redirect messages from replica nodes back to the primary. See
+    /// https://valkey.io/commands/client-capa/
+    ///
+    /// This is a `ValkeyClient` feature only. Redirection is handling differently with
+    /// `ValkeyClusterClient`.
+    var enableClientRedirect: Bool
```
nilanshu-sharma: This needs to be of `public` scope.

Sources/Valkey/ValkeyClientConfiguration.swift
```diff
@@ -248,6 +265,8 @@ public struct ValkeyClientConfiguration: Sendable {
     ///   - tls: The TLS configuration.
     ///   - databaseNumber: The Valkey Database number.
     ///   - readOnlyCommandNodeSelection: How we choose a node when processing readonly commands
+    ///   - enableClientRedirect: Support client redirection errors from replicas
+    ///   - redirectToPrimary: Redirect to primary node immediately if we are connected to a replica
```
nilanshu-sharma: Do you intend to say `connectToReplica` here ?

Sources/Valkey/ValkeyClientConfiguration.swift
```diff
@@ -230,6 +230,23 @@ public struct ValkeyClientConfiguration: Sendable {
...
+    ///
+    /// This is a `ValkeyClient` feature only. Redirection is handling differently with
+    /// `ValkeyClusterClient`.
+    var enableClientRedirect: Bool
+
+    /// Flag that we are connecting to a Replica and shouldn't redirect to the primary unless
+    /// we call a non readonly command
+    ///
+    /// This is a `ValkeyClient` feature only.
+    public var connectToReplica: Bool
```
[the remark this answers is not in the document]
adam-fowler: I have expanded the comment to add detail

adam-fowler:
> Looks good to me with a minor config rename suggestion.

I responded if you still feel we should change I can in another PR

PR #323, opened 2026-02-06
Cluster: Retry command on connection circuit breaker error

adam-fowler:
Retry sending command on connection circuit breaker error. In the case of single commands and transactions this will mean we ask for the cluster for a new node and send the command to that node. As long as the cluster has updated we should get a valid node.

Unfortunately with pipelined commands the code would need a major restructure if we were to provide a new node, so we just throw the circuit breaker error. I would hope to fix this at a later date.


nilanshu-sharma (commented):
Added a comment, free to merge after addressing these.

PR #329, opened 2026-02-11
Fix shard failover logic 

nilanshu-sharma:
The PR aims to cover corner cases identified during shard failover:  

1. During cluster failover, nodes report health status as "fail" in CLUSTER SHARDS responses, but the parser expects "failed". This causes:

  Error: "Invalid Node Health String: fail"

  This breaks cluster topology discovery during failover, preventing circuit breaker recovery and making the client permanently unavailable.

Ref: https://github.com/valkey-io/valkey/blob/unstable/src/cluster_legacy.c#L7065

```
<snip>:<snip>> cluster shards 
1) 1) "slots"
   2) 1) (integer) 10923
      2) (integer) 16383
   3) "nodes"
   4) 1)  1) "id"
          2) "9f8b5197a4222ac2e7ea30d892e5ee45bf39a7d9"
          3) "tls-port"
          4) (integer) <snip>
          5) "ip"
          6) "<snip>"
          7) "endpoint"
          8) "<snip>"
          9) "role"
         10) "replica"
         11) "replication-offset"
         12) (integer) 1766974
         13) "health"
         14) "online"
      2)  1) "id"
          2) "19b6336c741b8dd837fa0c759afe33ca9495610c"
          3) "tls-port"
          4) (integer) <snip>
          5) "ip"
          6) "<snip>"
          7) "endpoint"
          8) "<snip>"
          9) "role"
         10) "master"
         11) "replication-offset"
         12) (integer) 1766960
         13) "health"
         14) "fail"    <<------------------ fail instead of failed
```

2. Improve primary selection logic during cluster failover to handle split-brain scenarios gracefully. Instead of rejecting topologies with multiple primaries (which occurs temporarily during failover when the old primary is in state `fail` while the newly elected primary is `online`) . The logic prefer the online primary while falling back to any primary if none are online. This allows the cluster client to maintain availability during failover transitions instead of circuit-breaking.

### DIFF
--- a/Sources/Valkey/ValkeyClientError.swift
+++ b/Sources/Valkey/ValkeyClientError.swift
@@ -94,19 +94,42 @@ public struct ValkeyClientError: Error, CustomStringConvertible {
     public let message: String?
     /// If there is an underlying error it will be stored here
     public let underlyingError: Error?
+    /// Source file where the error was created
+    public let file: StaticString
+    /// Source line where the error was created
+    public let line: UInt
 
     /// Create a new error code.
     /// - Parameters:
     ///   - errorCode: The error code.
     ///   - message: The message to include.
-    public init(_ errorCode: ErrorCode, message: String? = nil, error: Error? = nil) {
+    ///   - error: The underlying error.
+    ///   - file: Source file (automatically captured).
+    ///   - line: Source line (automatically captured).
+    public init(
+        _ errorCode: ErrorCode,
+        message: String? = nil,
+        error: Error? = nil,
+        file: StaticString = #fileID,
+        line: UInt = #line
+    ) {
         self.errorCode = errorCode
         self.message = message
         self.underlyingError = error
+        self.file = file
+        self.line = line
     }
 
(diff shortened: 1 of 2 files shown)


Sources/Valkey/Cluster/ValkeyTopologyCandidate.swift
```diff
@@ -61,21 +61,31 @@ package struct ValkeyTopologyCandidate: Hashable {
 
         self.shards = try description.shards.map({ shard throws(ValkeyClusterError) in
             var primary: Node?
+            var firstPrimary: Node?
```
fabianfett: Can you add a code comment what the difference is between `primary` and `firstPrimary`?

nilanshu-sharma: I have refactored the code a bit to make it cleaner.

nilanshu-sharma:
Failover tests pointing to this branch look good. Please review when you get a chance @adam-fowler @fabianfett.

PR #368, opened 2026-03-09
Add valkey client cli application

adam-fowler:
To avoid having to keep hacking tests to verify command output, I've added a small CLI application for communicating with Valkey

### DIFF
--- a/Sources/ValkeyCLI/ValkeyCLI.swift
+++ b/Sources/ValkeyCLI/ValkeyCLI.swift
@@ -24,16 +24,14 @@ import Musl
 @main
 struct App {
     static func usage() -> Never {
-        print("Usage: ValkeyCLI <host/ip> <port>")
+        print("Usage: ValkeyCLI <host/ip>:<port>")
         exit(-1)
     }
 
     static func main() async throws {
-        let arguments = CommandLine.arguments
-        guard arguments.count > 2 else { usage() }
-        guard let port = Int(arguments[2]) else { usage() }
-        print("Connecting to \(arguments[1]):\(port)")
-        let valkeyClient = ValkeyClient(.hostname(arguments[1], port: port), logger: Logger(label: "ValkeyCLI"))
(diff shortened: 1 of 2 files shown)


Sources/ValkeyCLI/ValkeyCLI.swift
```diff
@@ -0,0 +1,84 @@
...
+        async let _ = valkeyClient.run()
+        while true {
+            print("> ", terminator: "")
+            guard let input = readLine() else {
+                print("")
+                return
+            }
+            let split = input.split(separator: " ")
+            guard let commandName = split.first else { continue }
+            let command = ValkeyRawCommand(String(commandName), parameters: split.dropFirst().map { String($0) })
```
nilanshu-sharma: Just a caveat worth calling out. This will not handle parameters which contain space character - `set key "hello world"`, but probably we don't need it here.

nilanshu-sharma (approved):
Looks good, just some minor comments

Sources/ValkeyCLI/ValkeyCLI.swift
```diff
@@ -70,6 +72,49 @@ struct App {
...
+                    if position != prevPosition {
+                        split.append(self[prevPosition..<position])
+                    }
+                    position = self.index(after: position)
+                    prevPosition = position
+                    continue
+                } else {
+                    insideSpeechMarks = false
+                    // we also consider speech marks as a separator
+                    split.append(self[prevPosition..<position])
```
nilanshu-sharma: Do we intentionally want to allow subsequence of size 0 here since it's missing the `if position != prevPosition` check ? 
If not, the logic to parse separator, and opening/closing speech marks is exactly same and can be tackled in a single if condition. 
Functionally the code looks good to me.

nilanshu-sharma (commented):
Mostly looks good, added a comment to potentially simplify the logic.

nilanshu-sharma (approved):

PR #369, opened 2026-03-10
Cluster-wide multi-key MGET 

nilanshu-sharma:
**Summary**

  - Add `ValkeyClusterMultiKeyCommand` protocol for commands that operate on keys spanning multiple hash slots
  - Implement cluster-wide `MGET` that transparently scatters sub-commands across nodes and reassembles results in original key order
  - Single-slot fast path skips scatter-gather and routes directly through the standard execute path
  - In future iterations this can be further optimized by opening per-node pipelines instead of per-slot pipeline.

### DIFF
--- a/Sources/Valkey/Cluster/ValkeyClusterClient+MultiKeyCommands.swift
+++ b/Sources/Valkey/Cluster/ValkeyClusterClient+MultiKeyCommands.swift
@@ -25,7 +25,7 @@ extension ValkeyClusterClient {
     ) async throws(ValkeyClientError) -> Command.Response {
         let keys = command.keysAffected
 
-        // Single-slot fast path
+        // Fast path for single slot or when there are no keys
         guard let partitions = partitionBySlot(keys: keys) else {
             return try await execute(command)
         }
@@ -33,10 +33,10 @@ extension ValkeyClusterClient {
         // Build one sub-command per slot
         let subCommands: [any ValkeyCommand] = partitions.map { command.createSubCommand(for: $0.indices) }
 
-        // Dispatch in parallel using the existing cross-node pipelining path.
+        // Dispatch in parallel using the existing cross-node pipelining path
         let rawResults = await self.execute(subCommands)
 
-        // Unwrap results and pair with original key indices for combineResults.
+        // Unwrap results and pair with original key indices for combineResults
         var slotResults: [(indices: [Int], result: RESPToken)] = []
         slotResults.reserveCapacity(partitions.count)
         for (i, partition) in partitions.enumerated() {
--- a/Sources/Valkey/Cluster/ValkeyClusterClient+MultiKeyCommands.swift
+++ b/Sources/Valkey/Cluster/ValkeyClusterClient+MultiKeyCommands.swift
@@ -1,0 +1,108 @@
+//
+// This source file is part of the valkey-swift project
+// Copyright (c) 2025 the valkey-swift project authors
+//
+// See LICENSE.txt for license information
+// SPDX-License-Identifier: Apache-2.0
+//
+
+@available(valkeySwift 1.0, *)
+extension ValkeyClusterClient {
+
+    // MARK: - Scatter-gather execute
+
+    /// Executes a multi-key command against the cluster using scatter-gather.
+    ///
+    /// Keys are partitioned by hash slot. One sub-command is dispatched per slot,
+    /// all slots are executed in parallel, and results are reassembled in the
+    /// original key order via ``ValkeyClusterMultiKeyCommand/assemble(originalKeyCount:slotResults:)``.
+    ///
+    /// - Parameter command: A command conforming to ``ValkeyClusterMultiKeyCommand``.
+    /// - Returns: The assembled command response in original key order.
+    /// - Throws: ``ValkeyClientError`` if execution on any node fails.
+    @inlinable
+    public func execute<Command: ValkeyClusterMultiKeyCommand>(
+        _ command: Command
+    ) async throws(ValkeyClientError) -> Command.Response {
+        let keys = command.keysAffected
+        let partitions = partitionBySlot(keys: keys)
+
+        // Build one sub-command per slot, cast to the type-erased protocol.
+        let subCommands: [any ValkeyCommand] = partitions.map { command.subCommand(for: $0.indices) }
+
+        // Dispatch in parallel using the existing cross-node pipelining path.
+        // Each sub-command contains only same-slot keys, so hashSlot() never
+        // throws keysInCommandRequireMultipleHashSlots for sub-commands.
+        let rawResults = await self.execute(subCommands)
+
+        // Map each raw Result<RESPToken, ValkeyClientError> to the typed Response.
+        let slotResults = partitions.enumerated().map { i, partition in
+            (indices: partition.indices, result: rawResults[i].convertFromRESP(to: Command.Response.self))
+        }
+
+        return try Command.assemble(originalKeyCount: keys.count, slotResults: slotResults)
+    }
+
+    // MARK: - Concrete mget / mset (shadow ValkeyClientProtocol extension methods)
+    //
+    // These concrete methods are preferred over the protocol-extension defaults
+    // for static dispatch on ValkeyClusterClient. They call execute(_:) which
+    // resolves — via static dispatch — to the constrained
+    // execute<Command: ValkeyClusterMultiKeyCommand> overload above, enabling
+    // transparent cross-slot scatter/gather.
+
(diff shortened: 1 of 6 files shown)


adam-fowler (commented):
I've had an initial look at this.
- There is a lot of memory allocation.
- The public protocol API needs some work. I'm not certain how it can be improved but if feels a bit hacky
- Have you considered how this is going to work for pipelined commands?

nilanshu-sharma:
@adam-fowler I have addressed your comments. Take a look when you get a chance. I am only implementing MGET here to keep the PR size manageable. If this looks good, we can add other popular multi-key commands like MSET etc.

Sources/Valkey/Cluster/ValkeyClusterClient+MultiKeyCommands.swift
```diff
@@ -0,0 +1,99 @@
...
+    /// Executes a multi-key command against the cluster using scatter-gather.
+    ///
+    /// Keys are partitioned by hash slot. One sub-command is dispatched per slot,
+    /// all slots are executed in parallel, and results are reassembled in the
+    /// original key order via ``ValkeyClusterMultiKeyCommand/assemble(originalKeyCount:slotResults:)``.
+    ///
+    /// - Parameter command: A command conforming to ``ValkeyClusterMultiKeyCommand``.
+    /// - Returns: The assembled command response in original key order.
+    /// - Throws: ``ValkeyClientError`` if execution on any node fails.
+    package func execute<Command: ValkeyClusterMultiKeyCommand>(
```
adam-fowler: Can we rename this function, so we don't get mixed up with standard `execute` function. Perhaps `executeMultiKeyCommand`.

Sources/Valkey/Cluster/ValkeyClusterClient+MultiKeyCommands.swift
```diff
@@ -0,0 +1,99 @@
...
+    /// Atomically returns the string values of one or more keys, transparently
+    /// routing sub-commands across cluster nodes for keys in different hash slots.
+    ///
+    /// - Documentation: [MGET](https://valkey.io/commands/mget)
+    /// - Complexity: O(N) where N is the number of keys to retrieve.
+    /// - Parameter keys: The keys whose values to retrieve.
+    /// - Returns: A ``RESPToken/Array`` with values in the same order as `keys`.
+    ///   Null tokens represent absent keys.
+    /// - Throws: ``ValkeyClientError`` if any node fails.
+    public func mget(keys: [ValkeyKey]) async throws(ValkeyClientError) -> RESPToken.Array {
```
adam-fowler: Move this to `ClusterStringCommands.swift` ?

Sources/Valkey/Cluster/ValkeyClusterClient+MultiKeyCommands.swift
```diff
@@ -0,0 +1,99 @@
...
+
+    // MARK: - Internal helpers
+
+    /// Executes a single command via the standard ``ValkeyCommand`` code path.
+    ///
+    /// This trampoline exists because `execute<C: ValkeyClusterMultiKeyCommand>`
+    /// would otherwise recursively call itself — Swift overload resolution prefers
+    /// the more constrained overload. Inside this helper the compiler only sees
+    /// `Command: ValkeyCommand`, so it resolves to the base `execute` overload.
+    private func executeSingleCommand<Command: ValkeyCommand>(
```
adam-fowler: Would need this if we used function name `executeMultiKeyCommand`

Sources/Valkey/Cluster/ValkeyClusterClient+MultiKeyCommands.swift
```diff
@@ -0,0 +1,99 @@
...
+    ) async throws(ValkeyClientError) -> Command.Response {
+        try await self.execute(command)
+    }
+
+    /// Groups key indices by hash slot.
+    ///
+    /// - Returns: An array of `(slot, indices)` pairs, one per unique slot.
+    private func partitionBySlot(
+        keys: some Collection<ValkeyKey>
+    ) -> [(slot: HashSlot, indices: [Int])] {
```
adam-fowler: Should this function return an optional array. If keys.count <= 1 return nil, if all keys belong to same slot don't create `slotIndices` array and return nil.

Tests/ValkeyTests/Cluster/ValkeyClusterClientTests.swift
```diff
@@ -247,6 +247,30 @@ actor TestCluster {
                 }
                 await self.setKey(key, value: value)
                 return .simpleString("OK")
+            case "MGET":
+                var keys: [String] = []
+                while let key = iterator.next() {
+                    keys.append(key)
+                }
+                var values: [RESP3Value] = []
+                for key in keys {
```
adam-fowler, line 277: Perhaps you should verify all keys are in the same hash slot. And return an error if they aren't

Sources/Valkey/Commands/Custom/ClusterStringCommands.swift
```diff
@@ -0,0 +1,62 @@
...
+
+            var resultIterator = resultArray.makeIterator()
+            for originalIndex in indices {
+                guard let element = resultIterator.next() else { break }
+                tokenBases[originalIndex] = element.base
+            }
+        }
+
+        // Build wire-format RESP3 array: `*N\r\n` followed by each element's bytes.
+        var buffer = ByteBuffer()
```
adam-fowler, line 51: You should be able to calculate the final size of the ByteBuffer and reserveCapacity

Sources/Valkey/Commands/Custom/ClusterStringCommands.swift
```diff
@@ -0,0 +1,62 @@
...
+    package static func combineResults(
+        originalKeyCount: Int,
+        slotResults: [(indices: [Int], result: Result<RESPToken, ValkeyClientError>)]
+    ) throws(ValkeyClientError) -> RESPToken.Array {
+        // Pre-fill every position with a RESP3 null token (`_\r\n`).
+        // Positions are overwritten with actual values as sub-results arrive.
+        var tokenBases = [ByteBuffer](repeating: RESPToken.nullToken.base, count: originalKeyCount)
+
+        for (indices, result) in slotResults {
+            let token = try result.get()
```
adam-fowler: You can replace the array conversion with 
```swift
let resultsArray = try RESPToken.Array(result.get())
```

Sources/Valkey/Commands/Custom/ClusterStringCommands.swift
```diff
@@ -0,0 +1,62 @@
...
+            let resultArray: RESPToken.Array
+            do {
+                resultArray = try RESPToken.Array(token)
+            } catch {
+                throw ValkeyClientError(.respDecodeError, error: error)
+            }
+
+            var resultIterator = resultArray.makeIterator()
+            for originalIndex in indices {
+                guard let element = resultIterator.next() else { break }
```
adam-fowler: Not getting a result is an error and we should throw an error perhaps `RESPDecodeError.invalidArraySize`

Sources/Valkey/Commands/Custom/ClusterStringCommands.swift
adam-fowler: Although you still have to catch the decode error and convert to ValkeyClientError. Perhaps combineResults should throw a `RESPDecodeError` and this is converted at a higher level

adam-fowler (commented):
More comments

Sources/Valkey/Commands/Custom/ClusterStringCommands.swift
nilanshu-sharma: Good idea!

Sources/Valkey/Cluster/ValkeyClusterClient+MultiKeyCommands.swift
nilanshu-sharma: Good idea - That would be a fast path that eliminates array allocations for the above scenario.

Sources/Valkey/Commands/Custom/ClusterStringCommands.swift
nilanshu-sharma: I have changed the implementation so that `combineResults()` throws `RESPDecodeError`, while the caller `executeMultiKeyCommand()` converting the error to `ValkeyClientError`.

nilanshu-sharma:
@adam-fowler Addressed then comments. Take a look when you get a chance.

adam-fowler (approved):
I think we are good

PR #373, opened 2026-04-03
Add internal sentinel client

adam-fowler:
This is a client to be used internally to communicate with sentinel nodes. It includes the following
- A topology election process similar to cluster
- Function to return current primary and replicas for a primary name


nilanshu-sharma:
@adam-fowler before I review this, quick question. How does the `ValkeySentinelClient` fit in the design here ? 
Do you intend to expose it to the users and they can use it to do their commands ? Or would it be used by another higher level client or ValkeyClusterClient ?

.devcontainer/docker-compose-replica.yml
```diff
@@ -18,14 +18,18 @@ services:
       - valkey_primary
```
nilanshu-sharma, line 18: Rename this to `docker-compose-sentinel.yml` ?

Sources/Valkey/Sentinel/ValkeySentinelClient.swift
```diff
@@ -0,0 +1,368 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+import Logging
+import NIOCore
+import NIOPosix
+import ServiceLifecycle
+import Synchronization
+
+@available(valkeySwift 1.0, *)
+package final class ValkeySentinelClient: Sendable {
```
nilanshu-sharma: Are you going to make it public later ? 
Will `ValkeySentinelClient` be used by users to run their commands ?

Sources/Valkey/Sentinel/ValkeySentinelClient.swift
```diff
@@ -0,0 +1,368 @@
...
+    let primaryName: String
+    let nodeClientFactory: ValkeyClusterNodeClientFactory
+    /// single node
+    @usableFromInline
+    let stateMachine: Mutex<StateMachine>
+    let logger: Logger
+    private let nodeDiscovery: any ValkeyNodeDiscovery
+
+    private enum RunAction {
+        case runSentinelDiscovery(runNodeDiscover: Bool)
```
nilanshu-sharma: Typo: `runNodeDiscovery`

Sources/Valkey/Sentinel/ValkeySentinelClient.swift
```diff
@@ -0,0 +1,368 @@
...
+                            } else {
+                                nil
+                            }
+                        }
+                    return (index, .init(primary: .hostname(primaryHost, port: primaryPort), replicas: replicas))
+                }
+            }
+            // Add a timeout task
+            group.addTask {
+                try await Task.sleep(for: .milliseconds(500))
```
nilanshu-sharma, line 103: Make this timeout configurable as part of `ValkeySentinelClientConfiguration.swift`

Sources/Valkey/Sentinel/ValkeySentinelClient.swift
```diff
@@ -0,0 +1,368 @@
...
+                        throw ValkeyClientError(.timeout)
+                    default:
+                        count -= 1
+                        self.logger.debug("Sentinel get primary address failed", metadata: ["error": .string("\(error)")])
+                        if count == 0 {
+                            throw ValkeySentinelError.sentinelIsUnavailable
+                        }
+                        break
+                    }
+                    break
```
nilanshu-sharma: These breaks are not needed.

Sources/Valkey/Sentinel/ValkeySentinelClient.swift
```diff
@@ -0,0 +1,368 @@
...
+        }
+    }
+
+    /// Get sentinel list returned by this sentinel
+    private func getSentinelList(_ node: ValkeyNodeClient, nodeID: ValkeyNodeID) async throws -> ValkeySentinelNodeList {
+        var sentinels = try await node.execute(SENTINEL.SENTINELS(primaryName: self.primaryName))
+            .decode(as: [SentinelInstance].self)
+            .compactMap {
+                // filter out disconnected or nodes flagged as `s_down`.
+                if !$0.flags.contains(.disconnected) && !$0.flags.contains(.s_down) {
```
nilanshu-sharma: We should also check objectively down nodes here as well

Sources/Valkey/Sentinel/ValkeySentinelError.swift
```diff
@@ -0,0 +1,27 @@
+//
+// This source file is part of the valkey-swift project
+// Copyright (c) 2025-2026 the valkey-swift project authors
+//
+// See LICENSE.txt for license information
+// SPDX-License-Identifier: Apache-2.0
+//
+
+/// Errors thrown from ValkeySentinelError
+public struct ValkeySentinelError: Error, Equatable {
```
nilanshu-sharma, line 10: Can we just use the existing ValkeyClientError instead of a new error struct for Sentinels.

Sources/Valkey/Sentinel/ValkeySentinelClientConfiguration.swift
```diff
@@ -0,0 +1,16 @@
+//
+// This source file is part of the valkey-swift project
+// Copyright (c) 2025-2026 the valkey-swift project authors
+//
+// See LICENSE.txt for license information
+// SPDX-License-Identifier: Apache-2.0
+//
+
+@available(valkeySwift 1.0, *)
+package struct ValkeySentinelClientConfiguration: Sendable {
```
nilanshu-sharma, line 10: If this is user facing this needs to be public.

Sources/Valkey/Sentinel/ValkeySentinelClient.swift
```diff
@@ -0,0 +1,368 @@
...
+                    }
+                    break
+                }
+            }
+            throw ValkeySentinelError.sentinelIsUnavailable
+        }
+    }
+
+    /// Subscribe to sentinel events
+    package func subscribeEvents(command: some ValkeySubscribeCommand, _ operation: (ValkeyClientSubscription) async throws -> Void) async throws {
```
nilanshu-sharma, line 129: Would this be used for the subscribing to events like `+switch-master` later on ?

nilanshu-sharma (changes requested):

adam-fowler:
> @adam-fowler before I review this, quick question. How does the `ValkeySentinelClient` fit in the design here ? 
> 
> Do you intend to expose it to the users and they can use it to do their commands ? Or would it be used by another higher level client or ValkeyClusterClient ? 

This will only ever be internal. To support sentinel a client needs to ask the sentinels what is the primary node every time a new connection is made. This is the code to get the primary node. 

Eventually the plan is to use this internally in ValkeyClient when connecting to Sentinel setups.

Sources/Valkey/Sentinel/ValkeySentinelClient.swift
adam-fowler: Yes

Sources/Valkey/Sentinel/ValkeySentinelError.swift
adam-fowler: The idea is you throw a `ValkeyClientError(.sentinelError, error: ValkeySentinelError.sentinelIsUnavailable)`

.devcontainer/docker-compose-replica.yml
adam-fowler: It used to test both client with replicas and sentinel configs

Sources/Valkey/Sentinel/ValkeySentinelClient.swift
adam-fowler: Not certain this should be configurable. Sentinel client spec expects sentinel to return within a "few hundreds of milliseconds". I have moved away from official spec to some degree in that I ask all the sentinels at the same time and the first one that returns a response is accepted. The spec suggests asking one at a time, if it doesn't respond within a "few hundreds of milliseconds" move onto the next one.

Sources/Valkey/Sentinel/ValkeySentinelClientConfiguration.swift
adam-fowler: not yet

nilanshu-sharma (approved):

PR #376, opened 2026-04-09
Fields type includes a count and an array. Replace with just an array

adam-fowler:
Currently a number of the hash commands include a `Fields` object that includes a count of fields and then the array of fields. The count is unnecessary and could cause bugs, as it could be out of sync with the array count. 

It would be preferable to call 
```swift
try await connection.hexpire(
  "myhash", 
  seconds: 10000, 
  fields: ["f2", "f3"]
)
```
over what we currently have
```swift
try await connection.hexpire(
  "myhash", 
  seconds: 10000, 
  fields: .init(numfields: 2, fields: ["f2", "f3"])
)
```
There are two ways to get the first call signature

- The first is to add `ExpressibleByArrayLiteral` to all the HCOMMAND.Fields types. This is a small change but, autocomplete will still recommend the old style API and it still allows users to create a Fields type where the `numfields` is different from the  actual fields count.
- The second is to patch the standard commands to replace the `Fields` type with `[Field]`. The new API would be recommended by autocomplete and it wouldn't be possible to setup an array with a count that is incorrect. This is a breaking change and would require some support for backwards compatibility. This is a bigger change.
 
In this PR you can see both alternatives in HashCustomCommands.swift. The first is in under the mark ["Additional API"](https://github.com/valkey-io/valkey-swift/pull/376/changes#diff-edf39d4839f93534822a8213849282bcfb6402a5473014c90cf27af83d292240R133), the second is under the mark ["Backwards Compatibility"](https://github.com/valkey-io/valkey-swift/pull/376/changes#diff-edf39d4839f93534822a8213849282bcfb6402a5473014c90cf27af83d292240R198).

### DIFF
--- a/Sources/Valkey/Commands/Custom/HashCustomCommands.swift
+++ b/Sources/Valkey/Commands/Custom/HashCustomCommands.swift
@@ -284,7 +284,14 @@ extension HSETEX {
             data.encode(into: &commandEncoder)
         }
     }
-    @available(*, deprecated, message: "Use init with `fields: [Field]` parameter")
+
+    @available(*, deprecated, renamed: "fieldsData")
+    var fields: Fields {
+        get { .init(numfields: self.fieldsData.count, data: self.fieldsData) }
+        set { self.fieldsData = newValue.data }
+    }
+
+    @available(*, deprecated, renamed: "hsetex(_:fieldsCondition:expiration:fieldsData:)")
     @inlinable
     public init(_ key: ValkeyKey, fieldsCondition: FieldsCondition? = nil, expiration: Expiration? = nil, fields: Fields) {
         self.key = key
@@ -463,7 +470,7 @@ extension ValkeyClientProtocol {
     ///     * 1: All the fields value and or expiration time was set.
     @inlinable
     @discardableResult
-    @available(*, deprecated, message: "Use version with `fields: [Field]` parameter")
+    @available(*, deprecated, renamed: "hsetex(_:fieldsCondition:expiration:fieldsData:)")
     public func hsetex<Field: RESPStringRenderable, Value: RESPStringRenderable>(
         _ key: ValkeyKey,
         fieldsCondition: HSETEX<Field, Value>.FieldsCondition? = nil,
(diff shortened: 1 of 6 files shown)


Sources/Valkey/Commands/Custom/HashCustomCommands.swift
```diff
@@ -127,3 +129,161 @@ extension HRANDFIELD {
...
+    ) async throws(ValkeyClientError) -> RESPToken.Array {
+        try await execute(HEXPIRE(key, seconds: seconds, condition: condition, fields: fields))
+    }
+}
+
+extension HEXPIREAT {
+    @available(*, deprecated, message: "Fields has been deprecated in favor of Array<Field>")
+    public typealias Fields = HashFields<Field>
+    @inlinable
+    @available(*, deprecated, message: "Use init with `field: [Field]` parameter")
```
nilanshu-sharma: Nit: Use init with `fields: [Field]`

Sources/Valkey/Commands/Custom/HashCustomCommands.swift
```diff
@@ -127,3 +129,161 @@ extension HRANDFIELD {
...
+    /// Set expiry time on hash fields.
+    ///
+    /// - Documentation: [HEXPIREAT](https://valkey.io/commands/hexpireat)
+    /// - Available: 9.0.0
+    /// - Complexity: O(N) where N is the number of specified fields.
+    /// - Returns: List of integer codes indicating the result of setting expiry on each specified field, in the same order as the fields are requested.
+    @inlinable
+    @discardableResult
+    @available(*, deprecated, message: "Use version with `field: [Field]` parameter")
+    public func hexpireat<Field: RESPStringRenderable>(
```
nilanshu-sharma, line 343: Nit: `hexpire` can be moved under this extension too so that all deprecated methods are in one place.

nilanshu-sharma (commented):
Just some minor comments. Approach 2 looks to be a better option here because it's going to flash warnings and autocomplete recommendations to point to the new API. 
It could be a considerable effort so you can also tackle the 2 commands in this PR and file an issue for the other ones if you want.

nilanshu-sharma (approved):

nilanshu-sharma:
Looks good to me!

PR #377, opened 2026-04-20
Fix continuation leak when idle-timer reschedules

ltetzlaff:
Running valkey-swift locally produces:

```
SWIFT TASK CONTINUATION MISUSE: runTimer(_:in:) leaked its
continuation without resuming it.
```

### Cause

When the idle-timeout timer fires while a lease request is queued (because a keep-alive is currently consuming the connection's only stream), `connectionIdleTimerTriggered` takes the `rescheduleIdleTimer` path. That path replaced the stale idle timer with a fresh one but dropped the old timer's `cancellationContinuation` on the floor.

The dropped token is a `CheckedContinuation<Void, Never>` that `ConnectionPool.runTimer`'s child task registered via `withCheckedContinuation`. Because the state machine holds it strongly via `State.Timer`, it is only deallocated when the state transition discards the timer — at which point it has never been resumed, and the runtime logs the warning above.

### Fix

- `ConnectionState.rescheduleIdleTimer()` and `ConnectionGroup.rescheduleIdleTimer(_:)` now also return the old idle timer's `TimerCancellationToken`.
- `PoolStateMachine.connectionIdleTimerTriggered(_:)` forwards that token via `.makeConnectionsCancelAndScheduleTimers` so `ConnectionPool.runConnectionAction` resumes the continuation before scheduling the new timer.

### Tests

A new `ConnectionPoolTests` target drives the state machine into the exact race (lease → establish → release → keep-alive fires → queue second request → idle timer fires while keep-alive is running) and asserts the returned action surfaces the old cancellation token. The test fails on `main` and passes with this fix.

Full existing suite (309 tests) continues to pass.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -129,6 +129,13 @@ var package = Package(
             ],
             swiftSettings: defaultSwiftSettings
         ),
+        .testTarget(
+            name: "ConnectionPoolTests",
+            dependencies: [
+                "_ValkeyConnectionPool"
+            ],
+            swiftSettings: defaultSwiftSettings
+        ),
     ]
 )
 
--- a/Sources/ValkeyConnectionPool/PoolStateMachine+ConnectionGroup.swift
+++ b/Sources/ValkeyConnectionPool/PoolStateMachine+ConnectionGroup.swift
@@ -585,7 +585,7 @@ extension PoolStateMachine {
         }
 
         @inlinable
-        mutating func rescheduleIdleTimer(_ connectionID: Connection.ID) -> ConnectionTimer? {
+        mutating func rescheduleIdleTimer(_ connectionID: Connection.ID) -> (ConnectionTimer, TimerCancellationToken?)? {
             guard let index = self.connections.firstIndex(where: { $0.id == connectionID }) else {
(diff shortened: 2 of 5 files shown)


adam-fowler:
Would you be able to do this fix in https://github.com/vapor/postgres-nio instead. The connection pool code is sourced from there and any fixes I've done in postgres-nio before implementing in valkey-swift.

ltetzlaff:
Sure thing — done: vapor/postgres-nio#641. Happy to keep this one open as a tracking/follow-up, or close it and reopen here once the postgres-nio side is merged and you pull the changes across, whichever you prefer.

adam-fowler:
> Sure thing — done: [vapor/postgres-nio#641](https://github.com/vapor/postgres-nio/pull/641). Happy to keep this one open as a tracking/follow-up, or close it and reopen here once the postgres-nio side is merged and you pull the changes across, whichever you prefer.

Once your fix is merged in postgres-nio. I'll copy across latest from postgres-nio and do a release

adam-fowler:
I'm going to close this in favour of #382 which includes this change

PR #378, opened 2026-04-22
Cluster-wide Multi-key MSET

nilanshu-sharma:
Extending Cluster-wide multi-key commands to also add MSET. 
MSET uses the existing multi-key command splitting and combining logic. 
Unit and Integration tests are added to verify the logic.

### DIFF
--- a/Sources/Valkey/Commands/Custom/ClusterStringCommands.swift
+++ b/Sources/Valkey/Commands/Custom/ClusterStringCommands.swift
@@ -69,9 +69,9 @@ extension MSET: ValkeyClusterMultiKeyCommand {
         MSET(data: indices.map { self.data[$0] })
     }
 
-    /// Combines per-slot MSET sub-results into a single `+OK` token.
+    /// Combines per-slot MSET sub-results into a single `OK` token.
     ///
-    /// Every sub-result must be the simple string `+OK`; any other response
+    /// Every sub-result must be the simple string `OK`. Any other response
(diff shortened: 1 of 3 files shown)


Sources/Valkey/Commands/Custom/ClusterStringCommands.swift
```diff
@@ -59,7 +59,38 @@ extension MGET: ValkeyClusterMultiKeyCommand {
...
+    /// Combines per-slot MSET sub-results into a single `OK` token.
+    ///
+    /// Every sub-result must be the simple string `OK`. Any other response
+    /// causes a ``RESPDecodeError`` to be thrown.
+    package static func combineResults(
+        originalKeyCount: Int,
+        slotResults: [(indices: [Int], result: RESPToken)]
+    ) throws(RESPDecodeError) -> RESPToken {
+        for (_, result) in slotResults {
+            guard result.base == okResponseBytes else {
```
adam-fowler, line 81: It might be nice to add a static to RESPToken
```swift
extension RESPToken {
    static var ok: Self { .init(validated: ByteBuffer(string: "+OK\r\n")) }
}
```
and then your test can be `guard result == .ok`. There are also a number of other places where this could be done as well. Search for the string `+OK\r\n`. This could be done in another PR as well

EDIT: In actual fact I just did it. #384

adam-fowler (approved):

PR #380, opened 2026-04-22
Fix compiler warnings in snippet

adam-fowler:


Snippets/MigratingFromRediStack.swift
```diff
@@ -28,7 +28,7 @@ func connectStandaloneExample() async throws {
         logger: logger
     )
 
-    try await withThrowingTaskGroup(of: Void.self) { group in
+    await withThrowingTaskGroup(of: Void.self) { group in
```
fabianfett: we should use a nonthrowing taskgroup here.

adam-fowler: This is kinda annoying really, in that an end user will almost be guaranteed to need a throwing task group, but yeah to be consistent I'll fix

nilanshu-sharma (approved):

PR #382, opened 2026-04-23
Latest changes from postgres-nio connection pool

adam-fowler:
- Support for draining connections
- Support for decreasing streams
- Continuation leak fix

### DIFF
--- a/Sources/ValkeyConnectionPool/ConnectionPool.swift
+++ b/Sources/ValkeyConnectionPool/ConnectionPool.swift
@@ -13,13 +13,13 @@ public struct ConnectionAndMetadata<Connection: PooledConnection>: Sendable {
 
 /// A connection that can be pooled in a ``ConnectionPool``
 public protocol PooledConnection: AnyObject, Sendable {
-    /// The connections identifier type.
+    /// The connection's identifier type.
     associatedtype ID: Hashable & Sendable
 
-    /// The connections identifier. The identifier is passed to
+    /// The connection's identifier. The identifier is passed to
     /// the connection factory method and must stay attached to
     /// the connection at all times. It must not change during
-    /// the connections lifetime.
+    /// the connection's lifetime.
     var id: ID { get }
 
     /// A method to register closures that are invoked when the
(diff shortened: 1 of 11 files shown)


Sources/ValkeyConnectionPool/ConnectionPool.swift
```diff
@@ -276,9 +276,11 @@ where
     }
 
     /// Mark a connection as going away. Connection implementors have to call this method if the connection
-    /// has received a close intent from the server. For example: an HTTP/2 GOWAY frame.
-    public func connectionWillClose(_ connection: Connection) {
-
+    /// has received a close intent from the server. For example: an HTTP/2 GOAWAY frame.
+    public func connectionWillClose(_ connectionID: ConnectionID) {
```
nilanshu-sharma, line 280: Make this non-public as it is in ConnectionPool ? Also we changed the signature of a public method, but I think this shouldn't be used directly by the users.

Sources/ValkeyConnectionPool/PoolStateMachine+ConnectionGroup.swift
```diff
@@ -384,6 +392,85 @@ extension PoolStateMachine {
...
+
+        @inlinable
+        mutating func connectionWillClose(_ connectionID: Connection.ID) -> ConnectionWillCloseAction {
+            guard let index = self.connections.firstIndex(where: { $0.id == connectionID }) else {
+                return .none
+            }
+
+            switch self.connections[index].markForClose() {
+            case .closeConnection(let closeAction):
+                self.stats.idle -= 1
```
nilanshu-sharma: This should should be an increment because after closing there will be 1 extra idle connection and 1 less leased connection ?

nilanshu-sharma (approved):
LGTM, just one comment around the public method signature change. Rest everything is from the upstream postgres-nio.

Sources/ValkeyConnectionPool/ConnectionPool.swift
adam-fowler: The other connection pool APIs are public as well. The module is not an exported module from valkey-swift so users can't access it anyway.

I would rather we keep changes as consistent with Postgres-nio as possible.

PR #386, opened 2026-04-28
Declare minimum platforms

ltetzlaff:
I was using `1.3.1` in a package and even though I was using  `platforms: [.macOS(.v15)]` it wouldn't work. 

Without explicit `platforms:` in the Package file SwiftPM seems to fallback to a default platform which for current toolchains seems to be macOS 10.13, which in turn makes the new `extension TaskGroup<Void>` and `extension DiscardingTaskGroup` from #382 fail to compile because TaskGroup, DiscardingTaskGroup and CheckedContinuation are only available on macOS 10.15+.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -27,6 +27,12 @@ defaultSwiftSettings.append(contentsOf: [
 
 var package = Package(
     name: "valkey-swift",
+    platforms: [
+        .macOS(.v13),
+        .iOS(.v16),
+        .tvOS(.v16),
+        .watchOS(.v9),
+    ],
     products: [
         .library(name: "Valkey", targets: ["Valkey"]),
         .library(name: "ValkeyBloom", targets: ["ValkeyBloom"]),


adam-fowler:
I'd rather we kept the platform requirements of the package as low as possible. Allows people to include valkey-swift as a dependency of a library within another package without requiring the platforms it requires for other libraries in that package (unless of course you use the one requiring valkey-swift). 

It would be preferable to add `@available` to the extensions. I'll upstream them to PostgresNIO after.

adam-fowler:
This should be fixed in 1.3.2

ltetzlaff:
Ah and it went unnoticed [there](https://github.com/vapor/postgres-nio/blame/820771e32c604de3342b82b05e15e824f37766de/Sources/ConnectionPoolModule/ConnectionPool.swift#L619) for so long because postgres-nio has platform requirements in place, gotcha!

I agree this is the preferable solution here 🙌

PR #388, opened 2026-05-15
Valkey 9.1 command changes

adam-fowler:
New commands
- CLUSTERSCAN (added custom response)
- HGETDEL
- MSETEX
New parameters
- HSETEX.keyCondition
- SCRIPTDEBUG.engine

Plus a whole load of documentation changes

9.1 is currently only available as RC so we shouldn't merge yet


nilanshu-sharma (approved):
Looks good. Approved. 
We should also add a valkey version compatibility matrix somewhere for every release.

PR #390, opened 2026-05-25
Only store hash value of healthyState of a cluster

adam-fowler:
The healthy state topology is only ever used to check if the topology has changed and if it has we update the hashSlotShardMap. This PR checks the hash value of the healthy state topology instead. This is a minor optimisation, but is also needed to remove the temptation of using the topology (which is not necessarily up to date) instead of the hashSlotShardMap.

### DIFF
--- a/Sources/Valkey/Cluster/ValkeyClusterClientStateMachine.swift
+++ b/Sources/Valkey/Cluster/ValkeyClusterClientStateMachine.swift
@@ -102,8 +102,6 @@ where
             @usableFromInline
             /* private */ var circuitBreakerTimer: Timer?
 
-            var lastHealthyState: ValkeyClusterTopology?
-
             var lastError: any Error
         }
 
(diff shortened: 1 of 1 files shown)


Sources/Valkey/Cluster/ValkeyClusterClientStateMachine.swift
```diff
@@ -281,18 +279,21 @@ where
...
-                    healthyState.clusterTopology
+                    healthyState.clusterTopologyHashValue
                 } else {
                     nil
                 }
             var result: ClusterDiscoverySucceededAction
+            let topologyHashValue = topology.hashValue
             /// Don't update cluster if description hasnt changed
-            if oldHealthyClusterTopology != topology {
+            if oldHealthyClusterTopologyHashValue != topology.hashValue {
```
nilanshu-sharma: You can use `topologyHashValue` here.

nilanshu-sharma (approved):

PR #391, opened 2026-05-25
Cluster readonly node AZ selection

adam-fowler:
- Add new readonly node selection method `az`. Which takes an availability zone and backup nodeSelection method. When the cluster is selecting a node for a readonly command it will select a node from the nodes in the availability zone. If no nodes are found in the zone it will use the backup selection method.
- This currently only works for Valkey 9.1 which returns the AZ in the `CLUSTER SHARDS` command. We can extend to earlier versions of Valkey by running the `INFO` command for every new node we find and extracting the AZ from there.
- I had to add a new Node type to store in the HashSlotShardMap as I didn't want to add the availability zone to the ValkeyNodeID. There are situations where we add nodes without knowing the AZ and will have to update it at a later date. So we don't want the AZ as part of an id.
- This is currently not supported by the standalone client, but should be pretty easy to extend.

### DIFF
--- a/Sources/Valkey/Cluster/ValkeyNodeID.swift
+++ b/Sources/Valkey/Cluster/ValkeyNodeID.swift
@@ -10,7 +10,7 @@
 /// `ValkeyNodeID` uniquely represents a node in the cluster by its network location
 /// (endpoint and port).
 @usableFromInline
-package struct ValkeyNodeID: Hashable, Sendable {
+package struct ValkeyNodeID: Hashable, Sendable, SelectableNodeIdentifier {
     /// The node's endpoint.
     ///
     /// This can be either a hostname (preferred) or an IP address.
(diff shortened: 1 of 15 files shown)


Sources/Valkey/Cluster/ValkeyNodeID.swift
```diff
@@ -26,6 +26,7 @@ package struct ValkeyNodeID: Hashable, Sendable {
     /// - Parameters:
     ///   - endpoint: The hostname or IP address of the Valkey node.
     ///   - port: The port number on which the node is accessible.
+    ///   - availabilityZone: Availability zone (used in node selection)
```
nilanshu-sharma: Minor: This doesn't match the initializer signature

Sources/Valkey/ValkeyClientConfiguration.swift
```diff
@@ -218,6 +219,10 @@ public struct ValkeyClientConfiguration: Sendable {
         public static var cycleReplicas: Self { .init(value: .cycleReplicas) }
         /// Cycle through primary and replicas
         public static var cycleAllNodes: Self { .init(value: .cycleAllNodes) }
+        /// Choose node based on availability zone, if no nodes are in availability zone use `backup`
+        public static func az(_ zone: String, backup: ReadOnlyCommandNodeSelection = .cycleReplicas) -> Self {
```
nilanshu-sharma, line 223: Lets guard against the case that users don't accidentally put backup node selection as `az` too.  With current approach users can nest az based node selections.

Sources/Valkey/ValkeyClientConfiguration+SwiftConfiguration.swift
```diff
@@ -46,14 +49,31 @@ extension ValkeyClientConfiguration {
...
+            self = .cycleReplicas
+        case "cycleAllNodes":
+            self = .cycleAllNodes
+        case .none:
+            guard let zone = configReader.string(forKey: ["readOnlyCommandNodeSelection", "availabilityZone"]) else {
+                self = .primary
+                return
             }
-            self.readOnlyCommandNodeSelection = .init(value: typed)
+            self = .az(zone)
```
nilanshu-sharma: This will set backup to `cycleReplicas` if someone uses Swift configuration.

nilanshu-sharma (changes requested):
Mostly looks good, just some minor comments around validations.

Sources/Valkey/ValkeyClientConfiguration+SwiftConfiguration.swift
adam-fowler: Added readOnlyCommandNodeSelection.availabilityZoneBackup

nilanshu-sharma (approved):

PR #392, opened 2026-06-16
Integrate swift-metrics

nilanshu-sharma:
Adds optional swift-metrics support behind a new MetricsSupport trait (default on), mirroring the existing DistributedTracingSupport pattern. Emits per-command latency, pipeline latency/size, and transaction latency/size through the globally bootstrapped MetricsSystem.

Following metrics are emitted as of now: 

- `valkey.command.<name>.duration` (Type = Timer, Dimension = Status {ok, error, timeout, cancelled})
- `valkey.pipeline.duration`  (Type = Timer, Dimension = n/a) 
- `valkey.pipeline.size`  (Type = Recorder, Dimension = n/a) 
- `valkey.transaction.duration`  (Type = Timer, Dimension = n/a) 
- `valkey.transaction.size`  (Type = Recorder, Dimension = n/a) 

Metric labels are fixed under the `valkey.*` prefix. Emission is opt-in via config.metrics.enabled (default false) - when disabled, the hot path skips clock reads and metric handler lookups entirely, so there is no overhead beyond a single comparison.


Sources/Valkey/ValkeyClientConfiguration.swift
```diff
@@ -306,6 +306,11 @@ public struct ValkeyClientConfiguration: Sendable {
     public var tracing: ValkeyTracingConfiguration = .init()
     #endif
 
+    #if MetricsSupport
+    /// The metrics configuration to use for the Valkey connection.
+    public var metrics: ValkeyMetricsConfiguration = .init()
+    #endif
+
```
adam-fowler, line 314: Should add swift-configuration support

Sources/Valkey/Connection/ValkeyConnection.swift
```diff
@@ -364,6 +397,12 @@ public final actor ValkeyConnection: ValkeyClientProtocol, Sendable {
         }
         #endif
 
+        #if MetricsSupport
```
adam-fowler: In some cases the pipeline metrics are already being called by the function calling `_execute`

adam-fowler (commented):
In general looks good. But there are some cases where the metrics are being recorded twice. I think we also need to expand the tests to include cluster retry attempts (with and without ask)

Sources/Valkey/ValkeyClientConfiguration.swift
nilanshu-sharma: Good catch, fixed it.

Sources/Valkey/Connection/ValkeyConnection.swift
nilanshu-sharma: Good Point - will move the metrics publishing from connection to the client layer - `ValkeyClusterClient` & `ValkeyClient`

adam-fowler:
I think metrics should be recorded at the connection level instead of client and cluster client. Currently if anyone calls a command from a connection no metric is recorded. But if you add recording to connection you will have to remove it from client and cluster client to avoid double recording everything.

nilanshu-sharma:
> I think metrics should be recorded at the connection level instead of client and cluster client. Currently if anyone calls a command from a connection no metric is recorded. But if you add recording to connection you will have to remove it from client and cluster client to avoid double recording everything.

I had the metrics getting emitted at connection level initially but then we were running into issues like multiple recordings of the same command for retry attempts / redirections etc. So I moved this logic to higher layers - `ValkeyClusterClient` and `ValkeyClient` for the following reasons: 

* We want to measure one logical command the user issued (valkey.command.*, valkey.pipeline.*, valkey.transaction.*), not one physical round-trip as there can be several round-trips per command. This also matches what users actually want: command latency, not per-attempt latency. Recording at the connection means a command that succeeds after a redirect emits 2–3 samples, inflating counts, skewing the latency histogram, and excluding the redirect/backoff wait the caller actually experiences.

* For pipelines and transactions, one batch fans out to N nodes (splitCommandsAcrossNodes), so you'd get N samples for one logical operation, each reporting only its slice - making the per-node latency and size metrics meaningless.

For commands run via `withConnection` don't emit metrics. That's intentional: it's the low-level API where the caller opts out of routing/retries/pooling. If we want connection-level visibility later, I'd add it as a separate metrics ( under `valkey.connection.*` ) rather than moving `valkey.command.*` down, so we don't conflate the two or double-count retries.

Tests/ValkeyTests/MetricsTests.swift
```diff
@@ -0,0 +1,471 @@
...
+                #expect(MetricsTests.factory.recorderSamples(label: "valkey.transaction.size") == [2.0])
+            }
+        }
+    }
+
+}
+
+// MARK: - In-memory metrics factory
+
+final class CapturingMetricsFactory: MetricsFactory, @unchecked Sendable {
```
adam-fowler: The metrics package already provides a debug metrics factory https://github.com/apple/swift-metrics/blob/main/Sources/MetricsTestKit/TestMetrics.swift. Perhaps we should use that.

Tests/ValkeyTests/MetricsTests.swift
```diff
@@ -0,0 +1,471 @@
...
+
+    @available(valkeySwift 1.0, *)
+    private func makeTopology() async -> TestStandaloneTopology {
+        await TestStandaloneTopology(primary: Self.primaryAddress, replicas: [])
+    }
+
+    @Test
+    @available(valkeySwift 1.0, *)
+    func testSingleCommandSuccessRecordsTimer() async throws {
+        Self.factory.reset()
```
adam-fowler: Instead of using a global metrics factory why don't you use a local one and use `withMetricsFactory`. You'll need to up the swift-metrics requirements to 2.9.0 for that
eg
```swift
let factory = CapturingMetricsFactory()
try await withMetricsFactory(factory) {
    try await withClient(...) { client in
        ...
    }
}
#expect(factory...)
```
You could even combine the two with functions into one `withClientAndMetricsFactory`. 
If you do this you can get rid of the `.serialized` parameter for the test suite.

Sources/Valkey/Connection/ValkeyConnectionFactory.swift
```diff
@@ -119,6 +119,13 @@ package final class ValkeyConnectionFactory: Sendable {
         #if DistributedTracingSupport
         var mConfig = newConfig
         mConfig.tracing = self.configuration.tracing
+        #if MetricsSupport
+        mConfig.metrics = self.configuration.metrics
+        #endif
+        return mConfig
+        #elseif MetricsSupport
+        var mConfig = newConfig
+        mConfig.metrics = self.configuration.metrics
```
adam-fowler: This might be cleaner
```swift
        #if DistributedTracingSupport || MetricsSupport
        var mConfig = newConfig
        #if DistributedTracingSupport
        mConfig.tracing = self.configuration.tracing
        #endif
        #if MetricsSupport
        mConfig.metrics = self.configuration.metrics
        #endif
        return mConfig
        #else
        return newConfig
        #endif
```

adam-fowler (commented):
A few more comments

Sources/Valkey/Connection/ValkeyConnectionConfiguration.swift
```diff
@@ -203,3 +213,18 @@ public struct ValkeyTracingConfiguration: Sendable {
...
+/// A configuration object that defines metrics emission behavior of a Valkey client.
+public struct ValkeyMetricsConfiguration: Sendable {
+    /// Whether metrics emission is enabled for this connection.
+    public var enabled: Bool
+
+    /// Create a new metrics configuration.
+    /// - Parameter enabled: Whether metrics emission is enabled. Defaults to `false`.
+    public init(enabled: Bool = false) {
+        self.enabled = enabled
+    }
```
fabianfett: this IMO should not be enabled but should be the actual MetricsFactory that you want to use. It should default to the global one.

Sources/Valkey/Metrics/ValkeyMetrics.swift
```diff
@@ -0,0 +1,188 @@
...
+/// command execution.
+@available(valkeySwift 1.0, *)
+@usableFromInline
+final class ValkeyCommandMetrics: Sendable {
+    @usableFromInline let ok: Timer
+    @usableFromInline let error: Timer
+    @usableFromInline let timeout: Timer
+    @usableFromInline let cancelled: Timer
+
+    init(commandName: String) {
```
fabianfett: please explicitly inject a metrics factory here.

Sources/Valkey/Metrics/ValkeyMetrics.swift
```diff
@@ -0,0 +1,188 @@
...
+}
+
+/// Pipeline-level metric pair: latency timer plus batch-size recorder.
+@available(valkeySwift 1.0, *)
+@usableFromInline
+struct ValkeyPipelineMetrics: Sendable {
+    @usableFromInline let timer: Timer
+    @usableFromInline let sizeRecorder: Recorder
+
+    init() {
```
fabianfett: please inject a metricsfactory here.

Sources/Valkey/Metrics/ValkeyMetrics.swift
```diff
@@ -0,0 +1,188 @@
...
+    }
+}
+
+/// Transaction-level metric pair: latency timer plus batch-size recorder.
+///
+/// Tracked separately from pipelines because MULTI/EXEC has different semantics and operators
+/// usually want to see transaction latency in isolation.
+@available(valkeySwift 1.0, *)
+@usableFromInline
+struct ValkeyTransactionMetrics: Sendable {
```
fabianfett: please inject a metrics factory here.

fabianfett (changes requested):
We should explicitly pass around the `MetricsFactory` here. Otherwise any kind of testing will become an absolute nightmare.

Also in tracing we explicilty allow overwriting all the names and labels. We should do the same here:
```
@available(valkeySwift 1.0, *)
/// A configuration object that defines distributed tracing behavior of a Valkey client.
public struct ValkeyTracingConfiguration: Sendable {
    /// The tracer to use, or `nil` to disable tracing.
    /// Defaults to the globally bootstrapped tracer.
    public var tracer: (any Tracer)? = InstrumentationSystem.tracer

    /// The attribute names used in spans created by Valkey. Defaults to OpenTelemetry semantics.
    public var attributeNames: AttributeNames = .init()

    /// The static attribute values used in spans created by Valkey.
    public var attributeValues: AttributeValues = .init()

    /// Attribute names used in spans created by Valkey.
    public struct AttributeNames: Sendable {
        public var databaseOperationName: String = "db.operation.name"
        public var databaseOperationBatchSize: String = "db.operation.batch.size"
        public var databaseSystemName: String = "db.system.name"
        public var networkPeerAddress: String = "network.peer.address"
        public var networkPeerPort: String = "network.peer.port"
        public var serverAddress: String = "server.address"
        public var serverPort: String = "server.port"
    }

    /// Static attribute values used in spans created by Valkey.
    public struct AttributeValues: Sendable {
        public var databaseSystem: String = "valkey"
    }
}
```

Please also include a Metrics.md that describes what metrics are collected and what their meaning is. Generally I think we should do the Metrics.md first and only then land the implementation.

Sources/Valkey/Metrics/ValkeyMetrics.swift
adam-fowler: @fabianfett have you seen the metrics task local variable. This PR could probably be updated to use that. 
```swift
withMetricsFactory(myMetrics) {
  doStuff()
}
```
Although because metrics are cached

adam-fowler: I guess that could also be done as a higher level as well. If you'd prefer to avoid the metrics factory lookup

fabianfett: I have seen that this API was added. I prefer passing the factory explicitly. The reason for this is simple. Having the option to grab the singleton or the task local at any time, leads devs to believe that metrics can be created  out of thin air at all the times. 

Hummingbird and Vapor I'm looking at you. ;) Passing the MetricsFactory around makes the pattern explicit. If the MetricsFactory is never stored and all Metrics are created at startup, further metrics can't be created later.

adam-fowler: This PR still creates metrics out of thin air, it just caches them (A pattern HB now follows). Are we wanting to change it so we create all metric primitives up front for all 400 valkey commands plus any extra from other libraries.

fabianfett: What is the benefit of having dedicated counters and timers for the different commands? Isn't this something that users potentially have to do anyway, as there is no way of knowing which GETs fail?

This is why I said, we likely need the proposal doc of what we want to count first.

adam-fowler: I was thinking the command could maybe be a dimension.
They possibly won't know which GET fails, but the more obscure commands they can probably recognise.

fabianfett: Yes. But all of this is very feelings-based – there is no story behind it. I'm not 💯 if I agree. Which is why we should first have a doc that explains which metrics we collect and why.

adam-fowler: Yeah a document would help. I believe @nilanshu-sharma is quite busy at the moment, so I'll write something up in a separate PR.

nilanshu-sharma:
Hey @adam-fowler @fabianfett . I have addressed a bunch of comments including: 
1) Passing in a metrics factory 
2) Using TestMetrics for tests 

This PR covers command metrics for simple commands as well pipelines and transactions (Duration & Size) at the ValkeyClient and ValkeyClusterClient layer which was my original intent for metrics support. But this PR is rather getting lengthy and I propose to build upon remaining features on top of it. Some of the things to be done as per comments : 
1) Custom labels to prefix the metric names 
2) Metrics at connection level - metric per retry   
3) Metrics coverage for commands run via `.withConnection()`

Sources/Valkey/Metrics/ValkeyMetrics.swift
```diff
@@ -0,0 +1,188 @@
...
+/// configuration.metrics.factory = myMetricsFactory
+/// ```
+public struct ValkeyMetricsConfiguration: Sendable {
+    /// The factory the client creates its metrics from, or `nil` to emit no metrics.
+    /// Defaults to `nil`.
+    ///
+    /// The client creates its metrics once, when it is initialized, and holds them for its lifetime.
+    /// Assigning `MetricsSystem.factory` therefore captures whichever factory is current at that
+    /// point, so `MetricsSystem.bootstrap(_:)` has to run first.
+    public var factory: (any MetricsFactory)?
```
adam-fowler, line 35: We need a public `init`

nilanshu-sharma: Good catch, fixed.

PR #394, opened 2026-07-02
Add support for setting the number of streams per connection.

adam-fowler:
Add option `ValkeyClientConfiguration.ConnectionPool.maximumNumberOfStreamsPerConnection ` to set number of streams a connection can support. This defaults to one, but increasing this will mean multiple processes can share connections and we can run with less open connections.

As this will mean we are sharing connections across processes this comes with a number of gotchas which are documented in the adjoining comment. Also if a command fails because a connection was closed due to cancellation it will retry it with a new connection.

### DIFF
--- a/Sources/Valkey/Node/ValkeyNodeClient.swift
+++ b/Sources/Valkey/Node/ValkeyNodeClient.swift
@@ -106,7 +106,7 @@ package final class ValkeyNodeClient: Sendable {
                 logger: logger
             )
 
-            return ConnectionAndMetadata(connection: connection, maximalStreamsOnConnection: 1)
+            return ConnectionAndMetadata(connection: connection, maximalStreamsOnConnection: connectionFactory.configuration.connectionPool.maximumNumberOfStreamsPerConnection)
         }
         self.connectionFactory = connectionFactory
         self.eventLoopGroup = eventLoopGroup
--- a/Sources/Valkey/Node/ValkeyNodeClient.swift
+++ b/Sources/Valkey/Node/ValkeyNodeClient.swift
@@ -86,6 +86,11 @@ package final class ValkeyNodeClient: Sendable {
         poolConfiguration.circuitBreakerTripAfter = connectionFactory.configuration.connectionPool.circuitBreakerTripAfter
         poolConfiguration.maximumConcurrentConnectionRequests = connectionFactory.configuration.connectionPool.maximumConcurrentConnectionRequests
 
+        precondition(
+            connectionFactory.configuration.connectionPool.maximumNumberOfStreamsPerConnection > 0,
+            "The maximum number of streams per connection must be greater than zero"
+        )
+
         self.readOnly = readOnly
         self.connectionPool = .init(
             configuration: poolConfiguration,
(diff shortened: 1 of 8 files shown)


Sources/Valkey/ValkeyClientConfiguration.swift
```diff
@@ -158,6 +158,22 @@ public struct ValkeyClientConfiguration: Sendable {
...
+        /// - Commands that take a long time can affect the performance of completely
+        ///   unrelated commands run elsewhere on the same connection.
+        /// - You cannot guarantee unique access to a connection. So transactions
+        ///   maybe affected by calls to `WATCH` elsewhere.
+        /// - If you are using seeing a large number of subscription push messages per second
+        ///   This could affect the performance of commands run on the same connection.
+        ///
+        /// In these situations you can start up a separate client with the maximum number of
+        /// streams set to one to run operations that are expensive or affect the state of a connection.
+        public var maximumNumberOfStreamsPerConnection: UInt16
```
nilanshu-sharma, line 175: We should guard against this value to be accidentally set to 0.

Sources/Valkey/ValkeyClientConfiguration.swift
```diff
@@ -199,6 +215,7 @@ public struct ValkeyClientConfiguration: Sendable {
             self.idleTimeout = idleTimeout
             self.circuitBreakerTripAfter = circuitBreakerTripAfter
             self.maximumConcurrentConnectionRequests = maximumConcurrentConnectionRequests
+            self.maximumNumberOfStreamsPerConnection = 1
```
nilanshu-sharma, line 218: Also also plumb this in SwiftConfiguration

nilanshu-sharma (changes requested):
Wow! didn't know enabling connection multiplexing is this easy with the swift connection pool. The logic looks good, added some minor comments.

Sources/Valkey/ValkeyClientConfiguration.swift
adam-fowler: done

Sources/Valkey/ValkeyClientConfiguration.swift
adam-fowler: Added precondition where it is used

nilanshu-sharma (approved):

Sources/Valkey/Node/ValkeyNodeClient.swift
```diff
@@ -106,7 +106,14 @@ package final class ValkeyNodeClient: Sendable {
                 logger: logger
             )
 
-            return ConnectionAndMetadata(connection: connection, maximalStreamsOnConnection: 1)
+            precondition(
+                connectionFactory.configuration.connectionPool.maximumNumberOfStreamsPerConnection > 0,
+                "The maximum number of streams per connection must be greater than zero"
+            )
```
fabianfett: move this before the closure so that we check this once before startup and not on every new connection :)

fabianfett (commented):
Generally this looks good to me however, I think we should have a unit test for this. Use a simple fake valkey handler and check, that we receive all the commands in the same handler.

adam-fowler:
> Generally this looks good to me however, I think we should have a unit test for this. Use a simple fake valkey handler and check, that we receive all the commands in the same handler.

I added a test using the MockConnections.

PR #396, opened 2026-07-02
Don't cancel other pending commands on command cancellation

adam-fowler:
This PR changes the affect of cancelling a command on the connection the command is running on.

Previously if a command was cancelled the connection the command was running on was closed and all other pending commands were failed with the error `.connectionClosedDueToCancellation`. The reasoning behind this was cancelling a command left the connection in an undefined state and we can't return a busy connection back to the connection pool.

This had one issue though. It meant that any other commands on that connection were also cancelled but without the knowledge of whether they actually occurred, which is problematic for non-idempotent commands like INCR.

This PR changes this to allow for pending commands to be processed, but also set the connection in the closing state, so once all the current pending commands have been processed the connection is closed and not returned to the connection pool.

### DIFF
--- a/Sources/Valkey/Connection/ValkeyChannelHandler+stateMachine.swift
+++ b/Sources/Valkey/Connection/ValkeyChannelHandler+stateMachine.swift
@@ -41,17 +41,16 @@ extension ValkeyChannelHandler {
             let context: Context
             var pendingCommands: Deque<PendingCommand>
 
-            func cancel(requestID: Int) -> (cancel: [PendingCommand], connectionClosedDueToCancellation: [PendingCommand]) {
-                var withRequestID = [PendingCommand]()
-                var withoutRequestID = [PendingCommand]()
-                for command in pendingCommands {
-                    if command.requestID == requestID {
-                        withRequestID.append(command)
-                    } else {
-                        withoutRequestID.append(command)
-                    }
+            mutating func cancel(requestID: Int) -> (cancel: [PendingCommand], stillPending: Bool) {
+                let index = pendingCommands.partition { $0.requestID == requestID }
+                let cancelled = [PendingCommand](pendingCommands[index...])
+                if index == pendingCommands.startIndex {
+                    pendingCommands.removeAll()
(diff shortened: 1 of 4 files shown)


Sources/Valkey/Connection/ValkeyChannelHandler+stateMachine.swift
```diff
@@ -41,17 +41,16 @@ extension ValkeyChannelHandler {
...
-                        withoutRequestID.append(command)
-                    }
+            mutating func cancel(requestID: Int) -> (cancel: [PendingCommand], stillPending: Bool) {
+                let index = pendingCommands.partition { $0.requestID == requestID }
+                let cancelled = [PendingCommand](pendingCommands[index...])
+                if index == pendingCommands.startIndex {
+                    pendingCommands.removeAll()
+                    return (cancelled, false)
+                } else {
+                    pendingCommands = .init(pendingCommands[..<index])
```
nilanshu-sharma: Since we altered the order of pending commands, can this change cause any potential ordering issues for the responses of commands which are not canceled ?

nilanshu-sharma (changes requested):

Sources/Valkey/Connection/ValkeyChannelHandler+stateMachine.swift
adam-fowler: oh yeah, I messed up there. Cancelling a command doesn't stop Valkey responding to it. I'll fix this

adam-fowler: This is now fixed. We now set cancelled commands to have a forget promise so when we receive a response for them it is ignored. We also drop any pending commands at the end of the pending command list whose response is to be ignored. This is so we move to the closed state as soon as we have dealt with all the commands that need to receive a response and don't hang around for responses we are only going to ignore.

I also split the cancel with pending test into two. One that cancels a command but then not another and one that cancels both commands.

nilanshu-sharma (approved):

PR #398, opened 2026-08-11
docs: add Pub/Sub section to README

edithturn:
The README covers Commands Execution, Pipelining, Transactions, and Connection Management for `ValkeyClient`, but has no Pub/Sub section. This PR adds a section under `ValkeyClient`, with a subscribe and a publish example:

It also documents a behaviour that is not mentioned anywhere: when you subscribe from a `ValkeyClient`, the subscription runs on its own dedicated connection, separate from the pool used for other commands. This means we can subscribe and publish on the same client without opening a second connection by ourselves.

I verified in two ways:

- In the source, `subscribe` to routes through [`_withSubscriptionConnection_`](https://github.com/valkey-io/valkey-swift/blob/f4a8915f1b8dcea303c4760babf130c6f896b5af/Sources/Valkey/Subscriptions/ValkeyClient%2Bsubscribe.swift#L18), which leases a dedicated connection via `leaseSubscriptionConnection`; INCR/PUBLISH use the normal pool.
- Running `CLIENT LIST` while connected showed two separate connections, one `cmd=subscribe` and one `cmd=publish`

Please let me know if this PR make any sense, and I appreciate any feedback.

Thanks, @stockholmux, for pointing out that this was worth documenting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,28 @@ try await valkeyClient.withConnection { connection in
 }
 ```
 
+#### Subscriptions (Pub/Sub)
+
+Subscribe to one or more channels and process messages as an async sequence. When the closure exits, the client automatically unsubscribes.
+
+```swift
+try await valkeyClient.subscribe(to: "notifications") { stream in
+    for try await message in stream {
+        print("Received: \(message.message)")
+    }
+}
+
+When you subscribe from a `ValkeyClient`, the subscription runs on a dedicated connection, kept separate from the connection pool used for every other command. Because of this, you can publish and run other commands on the *same* client while a subscription is active, the client manages the two connections for you.
+
+```swift
+// The client keeps subscribe and publish on separate connections automatically.
+try await valkeyClient.publish(channel: "notifications", message: "hello")
+```
+
+> **Note:** In Valkey (as in Redis), a connection in subscribe mode cannot issue most other commands, which traditionally means opening a second connection to publish. `ValkeyClient` handles this transparently by keeping the subscription on its own connection.
+
+
+
 ### ValkeyClusterClient
 
 For [Cluster Mode](https://valkey.io/topics/cluster-tutorial/) Valkey deployments with automatic command routing, topology discovery, and live cluster topology changes.


README.md
```diff
@@ -119,6 +119,28 @@ try await valkeyClient.withConnection { connection in
...
+}
+
+When you subscribe from a `ValkeyClient`, the subscription runs on a dedicated connection, kept separate from the connection pool used for every other command. Because of this, you can publish and run other commands on the *same* client while a subscription is active, the client manages the two connections for you.
+
+```swift
+// The client keeps subscribe and publish on separate connections automatically.
+try await valkeyClient.publish(channel: "notifications", message: "hello")
+```
+
+> **Note:** In Valkey (as in Redis), a connection in subscribe mode cannot issue most other commands, which traditionally means opening a second connection to publish. `ValkeyClient` handles this transparently by keeping the subscription on its own connection.
```
adam-fowler: This is only the case for the RESP2 protocol. Valkey-swift uses RESP3 which can run subscriptions and commands on the same connection. The reason for subscriptions running on a dedicated connection is they are long running and can hold a connection for a long time. Thus it is better to group them to run on one connection and not use multiple connections.

stockholmux: What @adam-fowler said, but I would replace 'traditionally' with 'in many other clients'

stockholmux (changes requested):
Seems like an improvement with a small change.

README.md
edithturn: Thanks @adam-fowler! You're right, it's RESP3. I've updated the Note to explain the dedicated connection as a design choice for long-running subscriptions. Happy to remove it entirely if you feel it adds more confusion than clarity, just let me know.

edithturn: Thanks  @stockholmux , I also updated that part.

README.md
```diff
@@ -119,6 +119,27 @@ try await valkeyClient.withConnection { connection in
 }
 ```
 
+#### Subscriptions (Pub/Sub)
+
+Subscribe to one or more channels and process messages as an async sequence. When the closure exits, the client automatically unsubscribes.
```
adam-fowler:
```suggestion
Subscribe to one or more channels and process messages as an AsyncSequence. When the closure exits, the client automatically unsubscribes.
```

adam-fowler (commented):
One minor thing, but otherwise looks good

README.md
edithturn: Thank you so much, Adam!

adam-fowler (approved):

PR #399, opened 2026-08-13
Add metrics proposal doc

adam-fowler:
This is a first stab at a proposal for the metrics valkey-swift should record

### DIFF
--- a/docs/metrics-proposal.md
+++ b/docs/metrics-proposal.md
@@ -1,0 +1,39 @@
+# Metrics proposal
+
+This document is a proposal for the metrics that valkey-swift should record
+
+The metrics labels included in this document should be configurable.
+
+## Command execution
+
+`valkey.command.duration`
+  - type: Timer
+  - dimensions: (status: [ok, error, timeout, cancelled], command: Command name)
+Record the time a command took to execute. Can be used to find out which commands are causing slowdown. Include status as a commands duration will be affected by these and could produce false positives. A command that returns an error immediately should not be grouped with a command that returns a result.
+
+`valkey.pipeline.duration`
+  - type: Timer
+  - dimensions: ?
+Record the time a pipeline took to execute. Unsure about dimensions as a pipeline of 2 commands will run a lot faster than a pipeline of 32 commands. But including the length as a dimension could be open to spamming the metrics backend.
+
+`valkey.transaction.duration`
+  - type: Timer
+  - dimensions: ?
+Record the time a transaction took to execute. Similar thoughts about dimensions
+
+## Connection Pool
+
+`valkey.connection-pool.connections`
+  - type: Meter
+  - dimensions: network address
+Number of open connections. Include network address as dimension to disambiguate between different nodes in a cluster.
+
+`valkey.connection-pool.idle`
+  - type: Meter
+  - dimensions: network address
+Number of idle connections. Include network address as dimension to disambiguate between different nodes in a cluster.
+
+`valkey.connection-pool.leased`
+  - type: Meter
+  - dimensions: network address
+Number of leased connections. Include network address as dimension to disambiguate between different nodes in a cluster.
--- a/docs/metrics-proposal.md
+++ b/docs/metrics-proposal.md
@@ -2,7 +2,7 @@
 
 This document is a proposal for the metrics that valkey-swift should record
 
-The metrics labels included in this document should be configurable.
+The metrics labels and dimensions included in this document should be configurable.
 
 ## Command execution
 
@@ -16,10 +16,18 @@ Record the time a command took to execute. Can be used to find out which command
   - dimensions: ?
 Record the time a pipeline took to execute. Unsure about dimensions as a pipeline of 2 commands will run a lot faster than a pipeline of 32 commands. But including the length as a dimension could be open to spamming the metrics backend.
 
+`valkey.pipeline.length`
+  - type: Recorder
+Record the length of a pipeline. Check for exceptionally long pipelines.
+
 `valkey.transaction.duration`
   - type: Timer
   - dimensions: ?
 Record the time a transaction took to execute. Similar thoughts about dimensions
+
+`valkey.transaction.length`
+  - type: Recorder
+Record the length of a transaction. Check for exceptionally long transaction.
 
 ## Connection Pool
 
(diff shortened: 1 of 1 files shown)


docs/metrics-proposal.md
```diff
@@ -0,0 +1,39 @@
...
+`valkey.pipeline.duration`
+  - type: Timer
+  - dimensions: ?
+Record the time a pipeline took to execute. Unsure about dimensions as a pipeline of 2 commands will run a lot faster than a pipeline of 32 commands. But including the length as a dimension could be open to spamming the metrics backend.
+
+`valkey.transaction.duration`
+  - type: Timer
+  - dimensions: ?
+Record the time a transaction took to execute. Similar thoughts about dimensions
+
```
adam-fowler, line 27: I haven't include the pipeline size recorder metrics as I don't see how they would be useful

docs/metrics-proposal.md
```diff
@@ -0,0 +1,39 @@
...
+## Connection Pool
+
```
adam-fowler, line 33: Connection Pool metrics are in here, but they should probably be discussed in postgres-nio as well

docs/metrics-proposal.md
```diff
@@ -0,0 +1,39 @@
+# Metrics proposal
+
+This document is a proposal for the metrics that valkey-swift should record
+
+The metrics labels included in this document should be configurable.
+
+## Command execution
+
+`valkey.command.duration`
+  - type: Timer
```
fabianfett, line 10: I think we will need this as a counter as well to know how many requests have succeeded, failed, ...

docs/metrics-proposal.md
```diff
@@ -0,0 +1,39 @@
...
+
+`valkey.connection-pool.idle`
+  - type: Meter
+  - dimensions: network address
+Number of idle connections. Include network address as dimension to disambiguate between different nodes in a cluster.
+
+`valkey.connection-pool.leased`
+  - type: Meter
+  - dimensions: network address
+Number of leased connections. Include network address as dimension to disambiguate between different nodes in a cluster.
```
fabianfett, line 57: this will also need: 

- connections_created
- connections_creation_failed
- connection_creation_time
- state: `running` vs `degraded` vs `circuitBreakerOpen`
- fulfilled_request
- request_time
- failed_requests

We should investigate what the correct pattern for total vs succeeded vs failed is. 3 counter? 2 counter (total and failed? success is computed = total - failed?)

docs/metrics-proposal.md
```diff
@@ -0,0 +1,39 @@
+# Metrics proposal
+
+This document is a proposal for the metrics that valkey-swift should record
+
+The metrics labels included in this document should be configurable.
```
fabianfett:
```suggestion
The metrics labels and dimension names included in this document should be configurable.
```

nilanshu-sharma: @fabianfett can you elaborate upon how labels and dimension can be configurable ? 
Do you intend to say prefix them with custom strings so that we can differentiate between metrics from multiple clients / apps ?

docs/metrics-proposal.md
nilanshu-sharma: I believe with timers we automatically get the counts so they serve dual-purpose here. And the dimension status {ok, canceled, timeout, error} will report the status.

docs/metrics-proposal.md
nilanshu-sharma: The pipeline & transaction size is important because sometimes the pipelines / transactions are slow because they are too big (I have seen users going crazy and doing 10s of thousands of operations in a pipeline). This way we can easily identify that the slowness is coming from pipelines which are too big and need to be split.

docs/metrics-proposal.md
```diff
@@ -0,0 +1,39 @@
...
+`valkey.connection-pool.connections`
+  - type: Meter
+  - dimensions: network address
+Number of open connections. Include network address as dimension to disambiguate between different nodes in a cluster.
+
+`valkey.connection-pool.idle`
+  - type: Meter
+  - dimensions: network address
+Number of idle connections. Include network address as dimension to disambiguate between different nodes in a cluster.
+
```
nilanshu-sharma, line 53: It's good to have visibility into pending and in-flight requests.

nilanshu-sharma (commented):
Besides the commands and connection. We should have metrics for: 
1) Topology Refresh via Cluster Shards
2) Circuit breaker on/off  
3)  ASK/MOVED redirections 
4) Command Retries

docs/metrics-proposal.md
adam-fowler: Timers do get counts. At least they do in Prometheus

docs/metrics-proposal.md
adam-fowler: @nilanshu-sharma In the same way `ValkeyTracingConfiguration` sets the attribute names

adam-fowler:
@fabianfett @nilanshu-sharma Perhaps we should be following the OTel conventions https://opentelemetry.io/docs/specs/semconv/db/database-metrics/
We can add additional metrics we feel are needed.

docs/metrics-proposal.md
fabianfett: sure. but this will make creating a dashboard way more involved. because users are required to do a sum over the buckets. other backends, may not do the counter. so if we go through swift-metrics, I'm pretty sure we need extra counters.

adam-fowler:
> @fabianfett @nilanshu-sharma Perhaps we should be following the OTel conventions https://opentelemetry.io/docs/specs/semconv/db/database-metrics/ We can add additional metrics we feel are needed.

bump

nilanshu-sharma:
@adam-fowler I didn't get notified for this thread, apologies. 
I am aligned toward having Otel conventions for the metrics which are listed in this [list](https://opentelemetry.io/docs/specs/semconv/db/database-metrics/) (such as `db.client.operation.duration`). 
For the metrics such as for pipelines, transactions which are not listed in OTel we can keep them completely custom `valkey.pipeline.*`, `valkey.transaction.*` so that they don't create a false illusion of being available in the convention list.

fabianfett:
@nilanshu-sharma there is rules for this to: https://opentelemetry.io/docs/specs/semconv/db/redis/

nilanshu-sharma:
Cool, yeah it has a metric name which can be used for both Pipeline and Transactions: [db.operation.batch.size](https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/) 
Also since this is for `redis`, I believe all of this is also applicable to `valkey` ?

adam-fowler:
> Cool, yeah it has a metric name which can be used for both Pipeline and Transactions: [db.operation.batch.size](https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/) Also since this is for `redis`, I believe all of this is also applicable to `valkey` ?

Also here is the section about pipelines and transactions
> [1] db.operation.name: It is RECOMMENDED to capture the value as provided by the application without attempting to do any case normalization. For [transactions and pipelined calls](https://redis.io/docs/latest/develop/clients/redis-py/transpipe/), if the individual operations are known to have the same command then that command SHOULD be used prepended by MULTI or PIPELINE . Otherwise db.operation.name SHOULD be MULTI or PIPELINE.

PR #401, opened 2026-08-16
Fix typo in channel handler state machine

fpseverino:
- Fixes a typo in the channel handler state machine, where a copy of the deque is created, and the `HELLO` command is prepended to it, but the original deque was then mistakenly returned with the action. Now the updated copy of the deque is returned.
- Adds a test that would fail without this fix


adam-fowler (approved):
Thanks for this @fpseverino

